PR #11, opened 2015-06-11
Add OPAM metadata files to permit direct installation of Infer

avsm:
This lets `opam pin add infer git://github.com/facebook/infer` work
out of the box by adding enough metadata to install all the
OCaml dependencies.  It currently doesn't install the Clang
analyzer in the build rules.

### DIFF
--- a/infer.install
+++ b/infer.install
@@ -1,0 +1,14 @@
+bin: [
+  "infer/bin/infer"
+  "infer/bin/inferTest"
+  "infer/bin/inferTraceBugs"
+  "infer/bin/InferJava"
+  "?infer/bin/InferClang"
+  "infer/bin/InferAnalyze"
+  "infer/bin/InferPrint"
+  "infer/bin/inferJ"
+  "infer/bin/BuckAnalyze"
+  "infer/bin/inferlib.py"
+  "infer/bin/utils.py"
+  "infer/bin/jwlib.py"
+]
--- a/opam
+++ b/opam
@@ -1,0 +1,23 @@
+opam-version: "1.2"
+name: "infer"
+version: "0.1.0"
+maintainer: "opam-devel@lists.ocaml.org"
+authors: "Facebook"
+homepage: "https://github.com/facebook/infer"
+bug-reports: "https://github.com/facebook/infer/issues"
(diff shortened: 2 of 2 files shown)


infer.install
```diff
@@ -0,0 +1,14 @@
+bin: [
+  "infer/bin/infer"
+  "infer/bin/inferTest"
+  "infer/bin/inferTraceBugs"
+  "infer/bin/InferJava"
+  "?infer/bin/InferClang"
```
jeremydubreil: Why the question mark at the beginning: `?infer/bin/InferClang` ?

infer.install
```diff
@@ -0,0 +1,14 @@
+bin: [
+  "infer/bin/infer"
```
jeremydubreil: The API of Infer is not very stable right now. So it is likely that these binaries are renamed, move of deleted. Is there a way to build this list automatically based on what is in infer/bin ?

opam
```diff
@@ -0,0 +1,23 @@
...
+depends: [
+  "ocamlfind" {build}
+  "sawja" {>="1.5"}
+  "atdgen" {>="1.5"}
+  "javalib" {>="2.3"}
+  "extlib" {>="1.5.4"}
+]
+depexts: [
+  [ ["ubuntu"] ["python2.7-dev"] ]
+  [ ["debian"] ["python2.7-dev"] ]
```
jeremydubreil: Will it be possible to support brew as well with this mechanism ?

infer.install
avsm: The question mark is for an optionally built file. The rest are all mandatory

opam
avsm: Sure -- what homebrew packages are needed to build it on OS X? Python works out of the box, so doesn't need to be specified.

infer.install
jeremydubreil: OK, thanks

opam
jeremydubreil: Nice. I was just asking if we could extend this file to support homebrew as me may need to in a short future.

jeremydubreil:
It seems that the merge did not work. We may have some idea about what failed. Would you mind sending the same pull request again ?

PR #13, opened 2015-06-11
Update infer to support symlinking

dambrisco:
While working on building a quick Homebrew formula for this, I ran into symlink-related issues around the libs folder. This seems to fix it with minimal impact.


akotulski:
Thank you for looking into this!

PR #33, opened 2015-06-12
Update INSTALL.md

qdaxb:
update osx intall path


jvillard:
Thanks for your PR! However that issue has already been fixed in 879d7757c75502c7c88bff515f9daac3da2b8e4a. This part of INSTALL.md was duplicating some info from our website, so we removed it.

Issue #26, opened 2015-06-12
ERROR when install infer from source on Mac OS X


fortonice:
infer (master) make -C infer      
make -C src clang
sed -e 's/@MAJOR@/0/g' \
		-e 's/@MINOR@/1/g' \
		-e 's/@PATCH@/0/g' \
		-e 's/@GIT_COMMIT@/2bce7c6c3dbb22646e2d67a2c6ade77f060b4bca/g' \
		-e 's/@GIT_BRANCH@/master/g' \
		-e 's/@GIT_TAG@//g' \
		backend/version.ml.in > backend/version.ml
atdgen -t backend/jsonbug.atd -o backend/jsonbug
make[1]: atdgen: No such file or directory
make[1]: *** [backend/jsonbug_t.ml] Error 1
make: *** [clang] Error 2

martinoluca:
Seems like you haven't installed some of the dependencies required to compile Infer.
Please look at the instructions provided for your OS: https://github.com/facebook/infer/blob/master/INSTALL.md#install-infer-from-source
and in particular to the point where you run:
`opam install sawja.1.5 atdgen.1.5.0 javalib.2.3 extlib.1.5.4`

tgkokk:
$ which -a atdgen
    /Users/tgkokk/.opam/4.01.0/bin/atdgen

It sounds like `opam` didn't install OCaml 4.01.0 correctly (or at all). Maybe you could run `opam init --comp=4.01.0`? Remember to answer `Y` to the question that asks you whether to update your shell's init file.

chris-wood:
I encountered this same issue. Below is the series of commands run, as per the installation instructions for OSX. Are you sure it's not something else?

$opam init --comp=4.01.0
$opam install sawja.1.5 atdgen.1.5.0 javalib.2.3 extlib.1.5.4
$./update-fcp.sh && ../facebook-clang-plugin/clang/setup.sh && ./compile-fcp.sh
$make -C infer
make -C src clang
sed -e 's/@MAJOR@/0/g' \
		-e 's/@MINOR@/1/g' \
		-e 's/@PATCH@/0/g' \
		-e 's/@GIT_COMMIT@/ee3f88a6c13f954aff8796c5536aa022b5c4bc20/g' \
		-e 's/@GIT_BRANCH@/master/g' \
		-e 's/@GIT_TAG@//g' \
		backend/version.ml.in > backend/version.ml
atdgen -t backend/jsonbug.atd -o backend/jsonbug
[...]

PR #40, opened 2015-06-12
OS X install script

marpaia:
A build script for building infer on a new OS X machine.

Help text:

```shell
$ ./scripts/mac_osx_build.sh --help
Usage: ./scripts/mac_osx_build.sh [options]

 options:
    --java-only   only build Java support (default: false)

 examples:
    ./scripts/mac_osx_build.sh                # build Java and C/Objective-C support
    ./scripts/mac_osx_build.sh  --java-only   # build Java support
```

I'm not sure if this is desired / needed, but I saw people having some trouble in #26, so I figured that I would document the installation steps in a reproducible tool.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,31 +46,40 @@ dependencies. Details below.
 
 Install the OCaml dependencies:
 
-```bash
-opam init --comp=4.01.0  # (answer 'y' to the question)
-eval `opam config env`
-opam install sawja.1.5 atdgen.1.5.0 javalib.2.3a extlib.1.5.4
+If you do not require support for the C/Objective-C analysis in Infer, and only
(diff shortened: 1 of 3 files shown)


dambrisco:
You'll probably want to set `--no-setup` on the `opam init` call if you want to keep environment clean. You can additionally set `OPAMROOT` to a temporary directory and add `$OPAMROOT/4.01.0/bin` to `PATH` to further prevent pollution (as opam isn't needed after install). 

There might be some other useful stuff in [this gist](https://gist.github.com/dambrisco/bd3009d2113e165532c0), but those are the most important bits. (Please note that the recipe in that gist doesn't work as-is, but the environment setup is correct.)

PR #44, opened 2015-06-13
Add OPAM metadata files to permit direct installation of Infer

avsm:
This lets `opam pin add git://github.com/facebook/infer` work
out of the box by adding enough metadata to install all the
OCaml dependencies.  It currently doesn't install the Clang
analyzer in the build rules.

### DIFF
--- a/infer.install
+++ b/infer.install
@@ -1,0 +1,14 @@
+bin: [
+  "infer/bin/infer"
+  "infer/bin/inferTest"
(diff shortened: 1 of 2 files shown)


avsm:
Weirdly, my `opam-files` branch didn't seem to exist for some reason, so I recreated it. I wonder if the merged failed because the originating branch disappeared.  I didn't delete it, but I'm not sure what else had the rights to do so...

PR #48, opened 2015-06-13
remove duplicated allocation

cs09g:
report_number is zero before the condition.
don't need to set zero again.

### DIFF
--- a/infer/bin/inferTraceBugs
+++ b/infer/bin/inferTraceBugs
@@ -187,9 +187,7 @@ class Selector(object):
         if len(self) > 1:
             report_number_str = raw_input(
                 'Choose report to display (default=0): ')
-            if report_number_str == '':
-                report_number = 0
-            else:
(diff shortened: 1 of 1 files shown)


infer/bin/inferTraceBugs
```diff
@@ -187,9 +187,7 @@ class Selector(object):
         if len(self) > 1:
             report_number_str = raw_input(
                 'Choose report to display (default=0): ')
-            if report_number_str == '':
-                report_number = 0
-            else:
+            if report_number_str != '':
                 report_number = self.parse_report_number(report_number_str)
         elif len(self) == 1:
             print('Auto-selecting the only report.')
```
jvillard: this branch also sets `report_number = 0` (on the line just below; can't click on it from github apparently...). Could you please remove that line as well and send a new pull request?

cs09g:
Sure. I'll send new one with that.

PR #72, opened 2015-06-16
remove duplicated allocation

cs09g:
report_number_str is already zero before the condition.


jvillard:
did you forget to make the change suggested in #48?

cs09g:
sorry, i missed. I'll do it again.

Issue #30, opened 2015-06-12
javalib incompatibility causes infer -- javac Hello.java to fail with no files analyzed


wywlds:
the log trace is as below:
wyw@wyw-System-Product-Name:~/workspace/test$ infer -- javac Hello.java 
TODO: print error message
Starting analysis (Infer version v0.1.0)
Fatal error: exception Sys_error("/home/wyw/workspace/test/infer-out/captured: No such file or directory")
[ERROR] Failure during create_makefile, original command was

[u'inferJ', u'-g', u'-a', 'infer', u'javac', 'Hello.java']

Traceback (most recent call last):
  File "/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/inferJ", line 19, in <module>
    stats = analysis.start()
  File "/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/inferlib.py", line 634, in start
    self.analyze_and_report()
  File "/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/inferlib.py", line 613, in analyze_and_report
    if self.analyze() == os.EX_OK:
  File "/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/inferlib.py", line 489, in analyze
    self.args.analyzer
  File "/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/inferlib.py", line 318, in run_command
    raise e
subprocess.CalledProcessError: Command '[u'/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/InferAnalyze', u'-results_dir', '/home/wyw/workspace/test/infer-out', u'-makefile', u'Makefile', u'-allow_specs_cleanup', u'-models', u'/home/wyw/Downloads/infer-linux64-v0.1.0/infer/infer/bin/../lib/java/models.jar', u'-project_root', '/home/wyw/workspace/test']' returned non-zero exit status 2

Anyone could explain?

martinoluca:
Hello @wywlds,
Have you tried running ` javac Hello.java` and see if it compiles correctly?
I suspect `javac` is not installed in your machine.

wywlds:
@martinoluca 
I installed javac.
Proof:
wyw@wyw-System-Product-Name:~/workspace/test$ javac -version
javac 1.7.0_06

jeremydubreil:
Thanks for reporting this. It seems that the execution reached the point where no source files were found:  
  
[...]

PR #94, opened 2015-06-23
[Java] Force javac to report in the correct language

occho:
This patch forces javac to report verbose output in English.

In my environment, the debug report in Japanese had caused Parsing.Parse_error at
https://github.com/facebook/infer/blob/master/infer/src/java/jClasspath.ml#L108.

This patch solved the error I encountered on compiling Hello.java.
I think `$ infer -- javac -J-Duser.language=ja Hello.java` will reproduce my bug if system supports the language. (My patch will not work with this case since the argument-specified language will override the language setting.)

Possible related issue: #30

### DIFF
--- a/infer/bin/jwlib.py
+++ b/infer/bin/jwlib.py
@@ -31,7 +31,8 @@ class CompilerCall:
         if self.args.version:
             return subprocess.call(['javac'] + self.original_arguments)
         else:
-            javac_cmd = ['javac', '-verbose', '-g', '-J-Duser.language=en'] + self.original_arguments
+            javac_cmd = ['javac', '-verbose', '-g'] + self.original_arguments
+            javac_cmd.append('-J-Duser.language=en')
 
             with tempfile.NamedTemporaryFile(
                     mode='w',
--- a/infer/bin/jwlib.py
+++ b/infer/bin/jwlib.py
@@ -31,7 +31,7 @@ class CompilerCall:
         if self.args.version:
             return subprocess.call(['javac'] + self.original_arguments)
         else:
-            javac_cmd = ['javac', '-verbose', '-g'] + self.original_arguments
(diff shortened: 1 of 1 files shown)


jeremydubreil:
@occho: Thanks for the pull request! We indeed did not consider this case.

infer/bin/jwlib.py
```diff
@@ -31,7 +31,7 @@ def run(self):
         if self.args.version:
             return subprocess.call(['javac'] + self.original_arguments)
         else:
-            javac_cmd = ['javac', '-verbose', '-g'] + self.original_arguments
+            javac_cmd = ['javac', '-verbose', '-g', '-J-Duser.language=en'] + self.original_arguments
```
jeremydubreil: I tried this with:

     `infer -- javac -verbose -J-Duser.language=ja Hello.java`

but it did not work for me because the command line then becomes:

  `['javac', '-verbose', '-g', '-J-Duser.language=en', '-verbose', '-J-Duser.language=ja', 'infer/bizarre/Test.java']`

So the second language option `'-J-Duser.language=ja'` is overriding the first one `'-J-Duser.language=en'`.

To make this work, you will need to add the local information at the end of the command line:

            javac_cmd = ['javac', '-verbose', '-g'] + self.original_arguments
            # force the verbose output to be in English
            javac_cmd.append('-J-Duser.language=en')

Will you be able to update the pull request to move the language option at the end ?

Thanks for this. It is good to make Infer more robust here because parsing the verbose output of javac is unfortunately a little bit fragile.

occho: Thank you for your review!
Updated as recommended.

jeremydubreil:
@occho: Thanks a lot for fixing this! It seems that this pull request is fixing one of the most frequently reported issue when running Infer on Java. We were unable to reproduce this bug because we only use english as default local language.

occho:
@jeremydubreil It's my pleasure! :)

PR #126, opened 2015-07-06
Exit with exit code 1 if there are errors with buck

tgummerer:
Summary:
Infer currently exits with exit code 0 whether the analysis returns
errors or not.  Return exit code 1 instead when there are errors, so
i.e. a CI server can interpret the results without parsing the output.

Test plan:
Run infer on a project with and without errors.

### DIFF
--- a/infer/bin/BuckAnalyze
+++ b/infer/bin/BuckAnalyze
@@ -273,12 +273,14 @@ def get_basic_stats(stats):
         'total_time',
     }
     bugs_found = 'Errors found:\n\n'
+    n_errors = 0
     for key, value in sorted(stats['int'].items()):
         if key not in to_skip:
+            n_errors += 1
             bugs_found += '  {0:>8}  {1}\n'.format(value, key)
 
     basic_stats_message = files_analyzed + phase_times + bugs_found + '\n'
-    return basic_stats_message
+    return basic_stats_message, n_errors
 
 
 def get_buck_stats():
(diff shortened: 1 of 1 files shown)


jvillard:
Hi @tgummerer, thanks for your diff! We're still in discussion about what to do and others have requested this feature as well (eg, #139).

We'll likely want to gate this behaviour behind a command-line option to Infer. Moreover, it would be good to implement this for other modules than just buck so that Infer behaves consistently on different build systems.

jeremydubreil:
There are cases where we do not want to exit with 1 when infer finds an error. A typical example is to add infer on the CI of an existing project. Then it will fail all the time until all the existing bugs are fixed which may take time.

A possible approach is to run Infer on the CI as it is, keeping the exit code 0 for a successful analysis, and create the exist code as a separate step:

  infer -o out  -- buck build target
  check_errors -o out/report.json

Adding a command line option is a possible approach but there are many different way in which Infer can be used on CI.

jeremydubreil:
Example of creating failures as an extra step: https://github.com/thyrlian/jooi

tgummerer:
@jvillard thanks for having a look!  Maybe it could be hidden behind a `--exit-code` flag.  I don't use infer with any of the other modules currently, but if there is interest in including this I could try to add to them as well.

@jeremydubreil That makes sense, I didn't think about that use case.  Thanks for the example. Personally I would prefer to avoid an extra script on the CI server though.

jvillard:
@tgummerer in the end I implemented something slightly more general that works for any infer module, not just buck: e63269fa14f360249612407d13cd47e35566ec48

Feel free to reopen this PR if needed.

tgummerer:
@jvillard thanks, I like this solution better than mine :)

PR #143, opened 2015-07-14
[java] Add TextUtils.isEmpty model

### LINKED ISSUE #141: [Java] False positive on Nullable
Infer doesn't call Android's TextUtils I guess?
```java
	public State(@Nullable String states) {
		if (!TextUtils.isEmpty(states)) {
			mSegments = states.split(";");
			trimSegments();
		} else {
			mSegments = new String[0];
		}
	}
```
Causes NULL_DEREFERENCE

denizt:
Add a partial copy of TextUtils from Android source for commonly used TextUtils.isEmpty method.

Fixes #141

### DIFF
--- a/infer/models/java/src/android/text/TextUtils.java
+++ b/infer/models/java/src/android/text/TextUtils.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2006 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jeremydubreil:
Would mind adding your example form https://github.com/facebook/infer/issues/141 in the list of tests: https://github.com/facebook/infer/blob/master/infer/tests/codetoanalyze/java/infer/NullPointerExceptions.java#L332

You should just need to add the example to the list of test code as the testing framework only matches the list of bugs that are found, which implies that the false positive has been fixed by adding your model.

PR #149, opened 2015-07-17
Indent json files

denizt:
The json files that were written by json.dump were valid, machine readable but they had no indentation/return lines making them a long single line and hard to read. Dump the json with an indent of
2.


jeremydubreil:
Thanks! This indeed makes the JSON output much more readable.

PR #188, opened 2015-10-15
Update create_binaries.sh

vampireshj2013:


jvillard:
@vampireshj2013 what's the issue you're trying to address? Maybe this would be better as a [GitHub issue](https://github.com/facebook/infer/issues/)?

jvillard:
Please re-open if I missed something.

PR #222, opened 2015-12-25
Add the opam modules required by build-infer

DavidAudrain:
While installing infer on MacOS X, build-infer.sh failed many times due to many missing opam modules.

The change specifies the list of opam modules that need to be installed in order to get build-infer.sh succeeding.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -47,6 +47,7 @@ release](http://fbinfer.com/docs/getting-started.html) (faster), or to
 compile everything from source (see the end of this document).
 
 - opam >= 1.2.0
+- Install opam modules `opam install ocamlfind atdgen camlzip extlib javalib sawja`
 - Python 2.7
 - Java (only needed for the Java analysis)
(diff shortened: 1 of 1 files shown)


jvillard:
Hi @DavidAudrain,

The idea here was that `./build-infer.sh` would tell you what to do in this situation: if the opam packages are missing, it fails with the following message:

```text
$ ./build-infer.sh
[ ... ]
checking for OCaml findlib package sawja >= 1.5.1... not found
configure: error: missing dependency: sawja.

If you are using opam, please run

  opam update
  opam pin add --yes --no-action .
  opam install --deps-only infer
```

Perhaps the last two commands can be added to the document instead of your opam command? The (small) advantage is that this is guaranteed to remain in sync with the [opam file](https://github.com/facebook/infer/blob/45813c3/opam). Also, please add the same commands to the MacOS instructions.

In the longer run, it will be possible to install infer directly from opam, which should also get rid of this issue, so an alternative solution is to just wait it out :wink:

cristianoc:
The opam file has been updated since.

PR #240, opened 2016-01-13
Don't create infer-out directory or logs unless analyzing something.

tonycosentini:
This will stop Infer from creating an `infer-out` directory when just running `infer`.

(I run `infer` without any arguments somewhat often to look at documenting.)

One downside to this is that there will be no logs when running Infer without any commands - is this bad?


jeremydubreil:
Thanks @tonycosentini  for the pull request! Yes, I think that it is legit to run `infer` as an equivalent to `infer --help`.

PR #258, opened 2016-02-01
Add android.app.Service to whitelisted Android initializer classes.

tonycosentini:
It has a similar lifecycle to activities and fragments, it would be nice to support `onCreate` as an initializer by default.


cristianoc:
Thanks for adding this!

Issue #270, opened 2016-02-09
Docker build fails


technion:
Hi,

I seem to have some issues with the docker build. Using the provided Dockerfile:

```
Step 10 : RUN cd /infer &&     eval $(opam config env) &&     make -C infer clang java
 ---> Running in 7d087081c6a2
[WARNING] Running as root is not recommended
make: Entering directory '/opt/infer-linux64-v0.6.0/infer'
make: Leaving directory '/opt/infer-linux64-v0.6.0/infer'
make: *** No rule to make target 'clang'.  Stop.
The command '/bin/sh -c cd /infer &&     eval $(opam config env) &&     make -C infer clang java' returned a non-zero code: 2
```

The beauty of Docker is that I don't think there's a lot of room for much to go wrong - any assistance appreciated.

kstreee:
It may works if you add './configure &&' command after the command 'eval $(opam config env) &&', but it also fails because of ocaml version.
The error message shown because of ocaml version is,
```
File "clang/cLocation.ml", line 76, characters 11-12:
Error: Syntax error
Command exited with code 2.
Makefile:187: recipe for target 'build_clang' failed
make[1]: *** [build_clang] Error 10
make[1]: Leaving directory '/opt/infer-linux64-v0.6.0/infer/src'
Makefile:21: recipe for target 'clang' failed
make: Leaving directory '/opt/infer-linux64-v0.6.0/infer'
make: *** [clang] Error 2
The command '/bin/sh -c cd /infer &&     eval $(opam config env) &&     ./configure &&     make -C infer clang java' returned a non-zero code: 2
```
And the build fail message comes from https://github.com/facebook/infer/blob/6389cb9bd0c47b21c72a2aa3532d0301ddb218fc/infer/src/clang/cLocation.ml#L76.
The syntax does not work with ocaml 4.01, and the ocaml installation command ( https://github.com/facebook/infer/blob/master/docker/Dockerfile#L20 ) installs ocaml 4.01.
Thus we need to change ocaml version using opam, ( opam switch 4.02.3 ).

I am currently trying to fix it and after I am done, I will send a pull request for it.

technion:
Thanks! I can confirm, I've managed to get that exact error out of this issue, so it does appear you've got it covered.

kstreee:
@technion I sent a pull request to solve it, so if you need it right now, you can try it. It works for me. https://github.com/facebook/infer/pull/271

jimmyxu0311:
[...]

PR #271, opened 2016-02-09
Solve failure in 'docker build .'.

kstreee:
An issue facebook/infer#270 suggest two docker build failure error.
facebok/infer#270 issues two docker build failure errors.
- failure in 'make -C infer clang java' : add './configure' before the command.
- failure in building sources : add 'opam switch 4.02.3' to use this syntax in ' https://github.com/kstreee/infer/blob/cc4439938e99a27d301a2d42aa3413849dc55a46/infer/src/clang/cLocation.ml#L76 '. In ubuntu, when 'apt-get install ocaml' ( https://github.com/kstreee/infer/blob/cc4439938e99a27d301a2d42aa3413849dc55a46/docker/Dockerfile#L20 ), for now, it installs ocaml 4.01, but this syntax does not work with ocaml 4.01.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,6 +52,7 @@ RUN INFER_VERSION=$(curl -s https://api.github.com/repos/facebook/infer/releases
 # Compile Infer
 RUN cd /infer && \
     eval $(opam config env) && \
+    ./configure && \
     make -C infer clang java
 
 # Install Infer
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN curl -sL \
       -o /usr/local/bin/opam && \
     chmod 755 /usr/local/bin/opam
 RUN opam init -y --comp=4.01.0 && \
+    opam switch 4.02.3 && \
```
jvillard: you can remove this line if you change the line above it to initialise the correct compiler:

```
  RUN opam init -y --comp=4.02.3 && \ 
```

kstreee:
Yes, you are right, it works after remove the 'opam switch 4.02.3' statement and replace --comp option value from 4.01.0 to 4.02.3. Thanks for your advice.

jvillard:
Thanks for working on this!

PR #272, opened 2016-02-09
build-infer.sh now works with non-traditional shells

### LINKED ISSUE #267: build-infer.sh: line 108: set: -g: invalid option
I am installing Infer from source following the install guide for MacOSX.

When I run `./build-infer.sh clang` or `./build-infer.sh java` I get:

```
initializing opam...
./build-infer.sh: line 108: set: -g: invalid option
set: usage: set [--abefhkmnptuvxBCHP] [-o option] [arg ...]
```

However, on line 108 in build-infer.sh you have `eval $(opam config env)`. I ran `eval $(opam config env)` and the command runs without errors.

If I just comment out line 108, `build-infer.sh clang` runs without errors. It just requests you run these commands before it will compile clang.

```
  opam update
  opam pin add --yes --no-action infer .
  opam install --deps-only
```

Does anyone else have this error? Any idea why you would get that error?

Regardless this is not a major issue. Just posting this in case someone else gets the same error.

Skyfold:
fixes #267.

By calling opam with the --shell=bash in the script I don't get the errors I had before.

The new line I added in the docker/README.md was so the list would display correctly when rendered.


build-infer.sh
```diff
@@ -105,7 +105,7 @@ check_installed opam
 opam init --no-setup --yes > /dev/null || \
   opam init --no-setup --yes --comp=4.02.3
 
-eval $(opam config env)
+eval $(opam config env --shell=bash)
```
jvillard: Unfortunately this doesn't work on opam 1.2.0 for instance. It seems this option was added between 1.2.0 and 1.2.2. I think this would achieve the same result:

```bash
eval $(SHELL=bash opam config env) 
```

PR #274, opened 2016-02-09
Make Java config test exit so it doesn't hang on some JVMs

kenkyee:
Was trying this article out and noticed it hung on the ./configure when trying to read JAVA_HOME:
https://medium.com/@Rapchik/static-analysis-build-server-for-android-with-jenkins-using-docker-bda888d4b34e#.dci57csez

Adding this fixed it...not sure why the JVM didn't just cleanly exit but apparently the openJDK JVM does this :-P


jvillard:
thanks @kenkyee!

Issue #283, opened 2016-02-18
Infer cannot find any issues when running in '/'


btakeya:
Hi all,
I think Infer cannot find procedure, when running in '/'.
In other directories(eg. /root, /usr/local, and so on), it finds procedure successfully.

Test case:
- Running in '/root'
```
root@416329cd0be2:~# pwd
/root
root@416329cd0be2:~# infer -- gcc /root/a.c 
Starting analysis (Infer version v0.7.0)
Computing dependencies... 100%
Analyzing 1 cluster. 100%
Analyzed 1 procedure in 1 file

Found 1 issue

a.c:6: error: NULL_DEREFERENCE
  pointer i last assigned on line 5 could be null and is dereferenced at line 6, column 3
  4.   {
  5.     int *i = NULL;
  6. >   *i = 3;
  7.     printf("%d\n", *i);
  8.   

Summary of the reports:

  NULL_DEREFERENCE: 1
```
- Result in '/root': 
```
root@416329cd0be2:~# cat /root/infer-out/proc_stats.json 
{ "files": 1, "procedures": 1, "lines": 10 }
```

--

- Running in '/'
```
root@416329cd0be2:~# cd /
root@416329cd0be2:/# infer -- gcc /root/a.c 
Starting analysis (Infer version v0.7.0)
Computing dependencies... 100%
Analyzing 0 clusters 100%
Analyzed 0 procedures in 1 file
No issues found
```
- Result in '/'
```
root@416329cd0be2:~# cat /infer-out/proc_stats.json 
{ "files": 1, "procedures": 0, "lines": 10 }
```

jvillard:
Thanks for the report @btakeya, I managed to reproduce the issue locally, we'll look into it.

btakeya:
this issue happens when running via docker (Infer version: v0.7.0)

kstreee:
This error comes from [```DB.source_file_equal (Exe_env.get_source exe_env pname) source_file```](https://github.com/facebook/infer/blob/master/infer/src/backend/inferanalyze.ml#L526).

The exact bug point is the function [```DB.rel_source_file_from_abs_path```]().
When the function with "/" parameter ```DB.rel_source_file_from_abs_path "/" "/root/a.c"```, then it returns ```Relative "/root/a.c"```.

I will explain how this bug make Infer skip analyzing procedures when run Infer on "/".

[...]

PR #284, opened 2016-02-20
[facebook/infer#283] Resolve a bug that Infer does not analyze procedures when the analyzer is run on '/'.

kstreee:
Resolve a bug that Infer does not analyze procedures when the analyzer is run on '/'.
This bug is reported by @btakeya on facebook/infer#283.


kstreee:
I am actually not sure that this is appropriate path to fix the bug, because this bug is caused by the function ```filename_to_relative``` that returns both of absolute path ('/' for the root parameter) and relative path. Should be replaced the function?

jvillard:
Thanks for looking into this, @kstreee! We'll be pushing further improvements to that part of the code soon as well.

Issue #288, opened 2016-02-26
Python import missing.


kstreee:
Following exception is raised for most current version of Infer,

```
Traceback (most recent call last):
  File "/home/skim/Documents/infer/infer/bin/infer", line 178, in <module>
    main()
  File "/home/skim/Documents/infer/infer/bin/infer", line 142, in main
    capture_exitcode = imported_module.gen_instance(args, cmd).capture()
  File "/home/skim/Documents/infer/infer/lib/python/inferlib/capture/mvn.py", line 78, in capture
    cmds = self.get_infer_commands(util.get_build_output(self.build_cmd))
  File "/home/skim/Documents/infer/infer/lib/python/inferlib/capture/mvn.py", line 48, in get_infer_commands
    capture = jwlib.create_infer_command(self.args, javac_args)
NameError: global name 'jwlib' is not defined
```

```from inferlib import jwlib``` is required because of [this line](https://github.com/facebook/infer/blob/9d080cadb19d02ee5ba45893ee03de113e7333e5/infer/lib/python/inferlib/capture/mvn.py#L48).

kstreee:
I think that Python code in Infer should have test code too.

PR #289, opened 2016-02-26
[facebook/infer#288] Fix missing imoprt statement bug.

kstreee:
Exception is raised because of missing import statment for ```jwlib```, facebook/infer#288.


jvillard:
Thanks @kstreee! maven integration is the one thing we don't have integration tests for internally, unfortunately. It would be cool to add some smoke tests on github to test that (easier to do it on the github side in this case).

PR #301, opened 2016-03-11
Add SuppressViewNullability annotation

NickFirmani:
This pull request adds the @SuppressViewNullability annotation. 

The reasoning behind this is that in libraries, one cannot use Butterknife for view binding, which forces you to do it manually. Basically, this makes a new annotation that infer treats the same way as @Bind/@InjectView

### DIFF
--- a/infer/annotations/com/facebook/infer/annotation/SuppressViewNullability.java
+++ b/infer/annotations/com/facebook/infer/annotation/SuppressViewNullability.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2004 - present Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.infer.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * View can be annotated with @SuppressViewNullability to silence warnings when
+ * a view is set to null in a destructor, and created in an initializer.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.FIELD)
+public @interface SuppressViewNullability {}
--- a/infer/src/checkers/annotations.ml
+++ b/infer/src/checkers/annotations.ml
@@ -140,12 +140,14 @@ let ia_is_initializer ia =
   ia_ends_with ia initializer_
 
 let ia_is_inject ia =
-  ia_ends_with ia inject
+  IList.exists
+    (ia_ends_with ia)
+    [inject; inject_view; bind; bind_array; bind_bitmap; bind_drawable; bind_string; suppress_view_nullability]
 
 let ia_is_inject_view ia =
   IList.exists
     (ia_ends_with ia)
-    [inject; inject_view; bind; bind_array; bind_bitmap; bind_drawable; bind_string; suppress_view_nullability]
+    [inject_view; bind; bind_array; bind_bitmap; bind_drawable; bind_string; suppress_view_nullability]
 
 let ia_is_mutable ia =
   ia_ends_with ia mutable_
--- a/infer/src/checkers/annotations.ml
+++ b/infer/src/checkers/annotations.ml
@@ -98,6 +98,10 @@ let initializer_ = "Initializer"
 let inject = "Inject"
 let inject_view = "InjectView"
 let bind = "Bind"
+let bind_array = "BindArray"
+let bind_bitmap = "BindBitmap"
+let bind_drawable = "BindDrawable"
+let bind_string = "BindString"
 let suppress_view_nullability = "SuppressViewNullability"
 let false_on_null = "FalseOnNull"
 let mutable_ = "Mutable"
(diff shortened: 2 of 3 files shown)


hzsweers:
Could definitely use this. We have an auxiliary `findViewById` method that's type-aware and returns exceptions when it's not found, so this would be immensely helpful.

cristianoc:
Is the idea to have an annotation with the same functionality of `@InjectView`:
  https://github.com/facebook/infer/blob/63b57790f91bcf8a89d6fc9e796f8b031915928b/infer/tests/codetoanalyze/java/eradicate/FieldNotNullable.java#L391
but with another name?

The functionality being: suppress both `FIELD_NOT_INITIALIZED` and `FIELD_NOT_NULLABLE`.
This pull request is an alias for `@Inject`, which suppresses only `FIELD_NOT_INITIALIZED`.
Just checking what's needed, as the description in the code indicates you'd like to suppress both.

hzsweers:
@cristianoc `@Inject` is also an alias for `FIELD_NOT_NULLABLE`, assuming being in the same group as the others implies it suppresses the same errors.

`[inject; inject_view; bind; suppress_view_nullability]`

cristianoc:
@hzsweers apologies for the code... that group is used to define annotations that have the property `ia_is_inject`, which suppresses `FIELD_NOT_NULLABLE`.
In addition `@InjectView` has the property `ia_is_inject_view`, which suppresses `FIELD_NOT_NULLABLE`.
So if you want both, you should modify the code for `ia_is_inject_view`.

Those properties should definitely be renamed to something less confusing: what they suppress.

hzsweers:
Ah interesting, should `ia_is_inject_view` be updated to include `bind` as well or is that left out since Butter Knife 7+ started using `Nullable` instead of `Optional`?

cristianoc:
@hzsweers I don't know the internal details of Butter Knife. If using it leads to `FIELD_NOT_NULLABLE` reports then we should include `bind` into `ia_is_inject_view` (or the better name for it).

hzsweers:
Basically:

Butter Knife 6 and before: `@InjectView` would cause a runtime exception if the view was null after binding. Could optionally add `@Optional` if you wanted to allow for nulls. Behavior applies to fields and methods.

```java
@InjectView(R.id.sample1) View myView;    // Cannot be null after bind(), will throw at runtime if it is
@Optional @InjectView(R.id.sample2) View myNullableView;    // Doesn't throw, can be null

@OnClick(R.id.blah1)
public void onClickBlah1() {
    // There must be a blah1 view, will throw at runtime if there's not
}

@Optional
@OnClick(R.id.blah2)
public void onClickBlah2() {
   // blah2 can be null, in which case this method can never be called
}
```

Butter Knife 7: `@Bind` was the same as `@InjectView` functionally, but `@Optional` was removed in favor of reusing `@Nullable` so the IDE would lint it. Same idea though. Behavior applies to fields and methods.

```java
@Bind(R.id.sample1) View myView;    // Cannot be null after bind(), will throw at runtime if it is
@Nullable @Bind(R.id.sample2) View myNullableView;    // Doesn't throw, can be null

@OnClick(R.id.blah1)
public void onClickBlah1() {
   // There must be a blah1 view, will throw at runtime if there's not
}

@Nullable    // This was weird, explained below
@OnClick(R.id.blah2)
public void onClickBlah2() {
  // blah2 can be null, in which case this method can never be called
}
```

Butter Knife 8+: Unreleased, but behavior will be the same with fields. For methods however, `@Optional` was re-added because it was strange to require `@Nullable` for methods since many had `void` return types.

```java
@Bind(R.id.sample1) View myView;    // Cannot be null after bind(), will throw at runtime if it is
@Nullable @Bind(R.id.sample2) View myNullableView;    // Doesn't throw, can be null

@OnClick(R.id.blah1)
public void onClickBlah1() {
   // There must be a blah1 view, will throw at runtime if there's not
}

@Optional
@OnClick(R.id.blah2)
public void onClickBlah2() {
  // blah2 can be null, in which case this method can never be called
}
```

---------

All in all: I think `bind` should be included as well then since its behavior is functionally the same. The problem arises in the situation where you have both `@Nullable` and `@Bind` annotations on a field, which I'm not sure if infer is aware and can ensure that `@Nullable` overrules.

cristianoc:
@hzsweers thanks for the detailed explanations!
Then definitely we should include `bind`. I'll make sure that `@Nullable` overrules `@Bind` if not already the case.

hzsweers:
No problem, happy to help!

For the scope of this PR, should @NickFirmani just move `suppress_view_nullability` to `ia_is_inject_view` and add `bind`?

Also there are some other new features in ButterKnife 7+ for resource value injection that should probably be there too, namely the following that bind reference types: 

```java
@BindDrawable
@BindBitmap
@BindArray
@BindString
```

Should they be added as well? They behave similarly in that they're null until `bind()` is called

cristianoc:
@hzsweers yes let's include them all.

NickFirmani:
Updated! Please look over for any goofs, don't really know Ocaml

hzsweers:
@NickFirmani I think you meant to add the butterknife annotations to `ia_is_injectview`, not `ia_is_inject`. I think `ia_is_inject` should probably just have `bind` and nothing else at this point, unless they need to be in both (@cristianoc correct me if I'm wrong)

Might also be a good time to give it a more general name, like `ia_is_butterknife_binding` or something

NickFirmani:
@hzsweers Thanks for the catch!

cristianoc:
Thank you for the pull request!

PR #302, opened 2016-03-14
First attempt to create a Checker

luiscruz:
@cristianoc 
This Checker is based on Checkers.callback_print_c_method_calls and prints every Java method call.

Note: for my convenience I have added ```oUnit-all.cache``` to ```.gitinore```.


infer/src/checkers/registerCheckers.ml
```diff
@@ -26,6 +26,7 @@ let active_procedure_checkers () =
         Checkers.callback_checkVisibleForTesting, false;
         Checkers.callback_check_write_to_parcel, false;
         Checkers.callback_find_deserialization, false;
+		Checkers.callback_print_java_method_call, false; (**danger luis*)
```
cristianoc: `false` --> `checkers_enabled`

sblackshear:
@luiscruz FYI,  we're rolling out a new framework that makes it easy to write a flow-sensitive checker if you can express your checker as an abstract interpretation. See https://github.com/facebook/infer/blob/dec843f5f53ced7a41bd3331de7be3c89a795b99/infer/src/checkers/copyPropagation.ml for an example of what these can look like.

The framework is restricted to intraprocedural checkers for now, but we're planning to extend it to summary-based interprocedural analyses in the near future.

luiscruz:
Thanks @sblackshear, it looks interesting. I'll take a look at that :+1:

Issue #299, opened 2016-03-11
How to implement custom checkers


luiscruz:
Hi there,

In the scope of my PhD I want to create a tool that performs some code analysis to detect common energy bugs on Android applications.
Since Infer is a bug detection tool I would like to write my analysis code on top of Infer.

As a proof of concept I want to create a simple checker:

- Analyse logging behaviour in the app.  

According to Moura et al. (2015), energy efficiency can be improved by disabling logging in the app.

For that I would need to:

1- detect standard logging function calls (such as System.out.println)
2- learn custom logging functions implemented by the developer
3- detect whether logging functions are being called

This does not look trivial to me, but as a start I would like to implement item (1) detect standard logging function calls.
If I’m able to do this I will have green light to proceed with other checkers.

I took a look into the source and I started studying Checkers.

I understand that checkers have to be registered in the src/checkers/registerCheckers.ml. E.g.:

```
ImmutableChecker.callback_check_immutable_cast, checkers_enabled;
```

So that java_checkers is a list like this:

```
[
	…;
	ImmutableChecker.callback_check_immutable_cast, checkers_enabled, Some Config.Java;
	…
]
```

I understand that the first argument (```ImmutableChecker.callback_check_immutable_cast```) is the function that will be called to perform this particular check.
This function points to another ```check_immutable_cast``` with the following arguments:

- curr_pname
- curr_pdesc
- typ_expected
- tap_found_opt
- loc 

These arguments are useful to check implicit type casts and were provided by ```Eradicate.callback_check_return_type```.
However if we need different information, I should process the following arguments:

```
{ Callbacks.get_proc_desc; proc_desc; proc_name; tend }
```

This is being done in ```callback_performance_checker```, but if I had some description of what are those variables and what can I do with them, it would make my life easier.
I see that there is some module to process procs descriptions (```Cfg.Procdesc.get_proc_name```), but I’m starting to get dizzy.

In addition, if you have any advice about the best way to implement this, I would be happy to hear about it.

#### Refs:
Moura, I., Pinto, G., Ebert, F., & Castor, F. (2015, May). Mining energy-aware commits. In Proceedings of the 12th Working Conference on Mining Software Repositories (pp. 56-67). IEEE Press.

cristianoc:
Simpler checkers to look at (they are just sample checkers which are disabled by default) are in `checkers.ml`.
In particular:
[...]

PR #303, opened 2016-03-16
added documentation explaining how to create a checker

luiscruz:
So this is the results of my first experience with Infer.
I have added a new document explaining how to create a Checker.
This is based on the thread https://github.com/facebook/infer/issues/299/.

I think this resource is useful in the official documentation since it will help developers to get started on how to contribute to Infer.
Feel free to change it or use it however you want.


cristianoc:
Thank you for the great document!

PR #374, opened 2016-05-27
Updating website template with newer design

matthewjohnston4:
As discussed with Jules, here is the newer version of the Open Source sites template, which is neater, better for mobile, and provides better information density.


matthewjohnston4:
Here's a before/after for desktop:

infer_compare_desktop

And another for mobile:

infer_compare_mobile

PR #429, opened 2016-08-05
Fixes/updates to build on linux.

ronin13:
a) An update to build-infer.sh to avoid downloading/building 4.02.3
ocaml compiler when it is already present.
b) Add the ifdef to avoid this error on glibc systems:

    clang -c -w libc_basic.c -o libc_basic.o
    libc_basic.c:692:12: error: unknown type name '__WAIT_STATUS'
    pid_t wait(__WAIT_STATUS stat_loc) {
            ^

### DIFF
--- a/infer/models/c/src/libc_basic.c
+++ b/infer/models/c/src/libc_basic.c
@@ -683,6 +683,16 @@ unsigned sleep(unsigned seconds) {
 #define __WAIT_STATUS int*
 #endif
 
+// glibc 2.24 did away with 'union wait' and replaced it with int*.
+// Hence, we re-define __WAIT_STATUS if glibc is >= 2.24.
+#ifndef __GLIBC_PREREQ
+# define __GLIBC_PREREQ(x,y) 0
+#endif
+
+#if defined(__GLIBC__) && __GLIBC_PREREQ(2, 24)
+#define __WAIT_STATUS int*
+#endif
+
 // the waiting is modeled as skip. Then the return value is a random value of a
 // process or
 // -1 in case of an error


facebook-github-bot-1:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

akotulski:
Regarding a)
We create separate opam switch intentionally to not pollute default opam switch (in case someone uses it for other ocaml development). The downside I know about is that ocaml needs to be rebuilt (but only once).

We introduced that logic to build-infer recently and we may revisit it in the future.

Regarding b)
Interesting find. Does compilation fail for you without it? What system and which version of gcc are you using?

ronin13:
> We create separate opam switch intentionally to not pollute default opam switch (in case someone uses it for other ocaml development). The downside I know about is that ocaml needs to be rebuilt (but only once).

Yes, even with this, it clones a separate compiler but it clones from existing one rather than downloading and building 4.02.3 compiler  again even if it is installed.

As you can see from https://gist.github.com/ronin13/5bd675296326c3bfdff9e2e3dfc26a88 it doesn't modify the opam switch.

> Regarding b)
> Interesting find. Does compilation fail for you without it? What system and which version of gcc are you using?

```
pacman -Q 'glibc' 'gcc-libs-multilib' 'gcc-multilib'
glibc 2.24-2
gcc-libs-multilib 6.1.1-5
gcc-multilib 6.1.1-5

lsb_release
LSB Version:    1.4

Distro is Arch Linux.
```


  and, yes, compilation fails with:
```

clang -c -w libc_basic.c -o libc_basic.o
libc_basic.c:692:12: error: unknown type name '__WAIT_STATUS'
pid_t wait(__WAIT_STATUS stat_loc) {
           ^
```

infer/models/c/src/libc_basic.c
```diff
@@ -682,6 +682,9 @@ unsigned sleep(unsigned seconds) {
 #ifdef __APPLE__ // define __WAIT_STATUS as int * on mac
 #define __WAIT_STATUS int*
 #endif
+#ifdef __GNUC__  // define __WAIT_STATUS as int * with glibc
```
akotulski: I would prefer to check for #ifndef __WAIT_STATUS instead of redefining __WAIT_STATUS (I think it gets defined on older platforms)

akotulski:
> Yes, even with this, it clones a separate compiler but it clones from existing one rather than downloading and building 4.02.3 compiler again even if it is installed.

I missed that part. Does it really work like this? I tried to reproduce what your pull request does on my mac and opam is still downloading and compiling ocaml with following script:
```
ALIAS_SWITCH=$(opam config var switch)

OPAMSWITCH="infer-4.02.3"
opam switch install -j 4 $OPAMSWITCH --alias-of $ALIAS_SWITCH || true
```

infer/models/c/src/libc_basic.c
ronin13: Cool, makes sense, will fix that. 

Would it better to do ifndef __WAIT_STATUS ... and then  ifdef  cygin/apple/gnuc .. endif?

akotulski: Actually I need to investigate it more. It looks like __WAIT_STATUS is a struct on some systems (headers that come with old gcc for example), please hold off with changes until I find out more

ronin13:
> ALIAS_SWITCH=$(opam config var switch)
> 
> OPAMSWITCH="infer-4.02.3"
> opam switch install -j 4 $OPAMSWITCH --alias-of $ALIAS_SWITCH || true

Regarding this, I believe (not an opam expert), it will download only if  `opam config var switch` returns a switch which doesn't map to 4.02.3 installed as the default.

So, for me, by default, 

```

opam config var ocaml-version
4.02.3

opam config var switch       
system

```
Full config: https://gist.github.com/ronin13/938747dbe8663f6dccf54f6c8e8096c6

jberdine:
Thanks @ronin13!  Could you split this into two separate pull requests since the two changes are not very closely related.  Also, could you change the comment in build-infer.sh to refer to $OCAML_VERSION instead of 4.02.3 so that it stays accurate as we upgrade.  And the comment would be better wrapped at 100 columns.

infer/models/c/src/libc_basic.c
ronin13: Thanks, will wait.

akotulski: Ok so it looks like `__WAIT_STATUS` got simplified into `int *` in newest glibc-2.24:
https://www.sourceware.org/ml/libc-alpha/2016-08/msg00212.html

You can check it with `__GLIBC_PREREQ(2, 24)` Example usage:
http://www.boost.org/doc/libs/1_33_1/boost/config/platform/linux.hpp

ronin13: Yes, seems like they got rid of the constant (__WAIT_STATUS, by replacing with int* everywhere) as well: https://github.com/bminor/glibc/commit/b49ab5f4503f36dcbf43f821f817da66b2931fe6 

I will use the ifdef, thanks.

ronin13:
@jberdine 

Yes, will do, will file a separate pull request for the ocaml one.

ronin13:
I have updated the pull request with only the build fixes. I will submit the build-infer.sh fix in a separate pull request as discussed earlier.

jvillard:
The PR seems to contain both changes still.

ronin13:
Sorry about that, fixed now.

jvillard:
@ronin13 thanks for the fix! If you can submit your other change as a new PR I'll be happy to merge it in as well.

ronin13:
@jvillard 

Thanks, will submit the other pull request soon.

ronin13:
@jvillard @jberdine  Submitted the pull request for build-infer here https://github.com/facebook/infer/pull/451

PR #437, opened 2016-08-11
Add documentation for some IR and backend modules

xennygrimmato:
- Brief description of modules and some functions provided by them

### DIFF
--- a/infer/src/IR/README.md
+++ b/infer/src/IR/README.md
@@ -13,11 +13,13 @@ The type environment module is [Tenv](Tenv.rei).
 ## Cg
 
 The Call Graph module ([Cg](Cg.re)) contains the following attributes in the Call Graph type:
+
   - Path to source file
   - Number of lines of code in source file
   - Map of Procedure Hash to Procedure Info (referred as Procname.Hash and node_info respectively)
 
 Procedure Info consists of the following attributes:
+
   - Parent Procedures
   - Children Procedures
   - Ancestors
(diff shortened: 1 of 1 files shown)


cristianoc:
@jberdine @jvillard do we have a Makefile target to generate ocamldoc from the code?
Just wondering what should go in .mli/.rei files directly vs in README.md files.

jberdine:
Last time I tried, ocamldoc fell over on infer.  It was a while ago though.  Still, the documentation comments can sometimes be shown by merlin (like show type, but show the doc comment) using merlin-document in emacs or however other editors expose it.  The doc comments also can be shown in the auto-completion pop-ups.  So I would say to take at least that perspective: put documentation at the level of single functions in doc comments in .mli/.rei, and higher-level or broader scope information could go elsewhere.  Although info that is specific to a file might best be a comment in that file.

xennygrimmato:
So, should I add comments to source files instead of Readme files to continue with this?

cristianoc:
@xennygrimmato yes I think that's better.
There's value, in the README, to explain what the concepts and parts are: there's a call graph, there's a control-flow graph, perhaps what their main function is, big picture kind of description. But for the details of each module, especially those details that are likely to need to be maintained when those modules evolve over time, it's best to have comments in the source files.

jberdine:

jeremydubreil:
Any update on this?

jvillard:
The implementation has diverged from the documentation in this PR so closing for now. Do feel free to re-open with an updated version.

PR #440, opened 2016-08-12
Remove explicit return of unit and use ignore instead

xennygrimmato:
Instead of returning `()` explicitly, use `ignore (f x)` to ignore whatever is returned by `f` and return `()`.


jvillard:
Please make this just one commit instead of four.

jberdine:

my3190:
master xennygrimmato

cristianoc:
The file was unused and has been deleted, closing this.

Issue #387, opened 2016-06-21
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe7 


LichFaker:
Hi, I type this into my terminal:
`infer -- gradle clean build`

However,  I got an error:
![image](https://cloud.githubusercontent.com/assets/5850082/16217793/5f879038-37a9-11e6-9293-ed88cf84b8e3.png)

I want to know why and how to solve it ?

chen881129:
infer-0.8.1/infer/lib/python/inferlib/capture/util.py: 27

`return verbose_out_chars.decode('gbk').split('\n')`
改成这样试试

jvillard:
This should be fixed in latest master from GitHub. We should make a release but in the meantime you can compile Infer from source.

LichFaker:
@jvillard Okay, thank u very much

loveai88:
I run in OSX ,and change it to verbose_out_chars.decode('UTF-8').split('\n').It is ok

LichFaker:
@loveai88  thanks

PR #443, opened 2016-08-24
Fix ascii codec can't decode utf8

Hackforid:
Fix Issue https://github.com/facebook/infer/issues/387

There is a error that python2.x not support ascii as default
so we got 
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe5 in position 2944712: ordinal not in range(128)
```


facebook-github-bot-3:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

PR #444, opened 2016-08-24
Fix ascii codec can't decode utf8

Hackforid:
Fix Issue https://github.com/facebook/infer/issues/387

There is a error that python2.x not support ascii as default
so we got 
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe5 in position 2944712: ordinal not in range(128)
```

### DIFF
--- a/infer/lib/python/inferlib/capture/util.py
+++ b/infer/lib/python/inferlib/capture/util.py
@@ -24,7 +24,7 @@ def get_build_output(build_cmd):
     #  TODO make it return generator to be able to handle large builds
     proc = subprocess.Popen(build_cmd, stdout=subprocess.PIPE)
     (verbose_out_chars, _) = proc.communicate()
-    return utils.decode(verbose_out_chars).split('\n')
+    return utils.decode(verbose_out_chars).decode('UTF-8').split('\n')
 
 
 def run_compilation_commands(cmds, clean_cmd):


facebook-github-bot-4:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-7:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

infer/lib/python/inferlib/capture/util.py
```diff
@@ -24,7 +24,7 @@ def get_build_output(build_cmd):
     #  TODO make it return generator to be able to handle large builds
     proc = subprocess.Popen(build_cmd, stdout=subprocess.PIPE)
     (verbose_out_chars, _) = proc.communicate()
-    return utils.decode(verbose_out_chars).split('\n')
+    return utils.decode(verbose_out_chars).decode('UTF-8').split('\n')
```
jvillard: `utils.decode()` will already call `decode()`, so I have trouble understanding what this is doing. Is there still an issue with the current version of infer? #387 refers to an outdated problem that was fixed (by calling `utils.decode()` here).

PR #451, opened 2016-08-31
Use the system compiler if it matches the supported one.

ronin13:
This avoids the need to download the ocaml compiler and building it,
if the system one is already  $SUPP_OCAML_VERSION which is 4.02.3 as of
now.

### DIFF
--- a/build-infer.sh
+++ b/build-infer.sh
@@ -39,8 +39,10 @@ BUILD_CLANG=no
 BUILD_JAVA=no
 INTERACTIVE=yes
 ORIG_ARGS="$*"
+# ocaml version that is supported
+SUPP_OCAML_VERSION="4.02.3"
 
-while [[ $# > 0 ]]; do
+while [[ $# -gt 0 ]]; do
   opt_key="$1"
   case $opt_key in
     all)
@@ -96,12 +98,21 @@ check_installed () {
 }
 
 setup_opam () {
-    OCAML_VERSION="4.02.3"
-
     opam init -j $NCPU --no-setup --yes
 
-    OPAMSWITCH=infer-$OCAML_VERSION
-    opam switch install -j $NCPU $OPAMSWITCH --alias-of $OCAML_VERSION || true
+    # get the present ocaml version, the idea here is to avoid installing ocaml compiler if the
+    # default is $SUPP_OCAML_VERSION.
+    present=$(opam config var ocaml-version)
+    if [[ $present == $SUPP_OCAML_VERSION ]];then
+        # get the switch corresponding the version
+        ALIAS_SWITCH=$(opam config var switch)
+    else
+        # if the default is not $SUPP_OCAML_VERSION, download ocaml compiler corresponding to it
+        ALIAS_SWITCH=$SUPP_OCAML_VERSION
+    fi
+
+    OPAMSWITCH=infer-$SUPP_OCAML_VERSION
+    opam switch install -j $NCPU $OPAMSWITCH --alias-of $ALIAS_SWITCH || true
 }
 
 add_opam_git_pin () {


jberdine:
Thanks!  Just for code style consistency, could you wrap the comments at closer to column 100, and change capitalization and punctuation to match the existing comments?

ronin13:
Done.

jvillard:
Thanks @ronin13!

@facebook-github-bot shipit

build-infer.sh
```diff
@@ -96,12 +98,21 @@ check_installed () {
...
+    if [[ $present == $SUPP_OCAML_VERSION ]];then
+        # get the switch corresponding the version
+        ALIAS_SWITCH=$(opam config var switch)
+    else
+        # if the default is not $SUPP_OCAML_VERSION, download ocaml compiler corresponding to it
+        ALIAS_SWITCH=$SUPP_OCAML_VERSION
+    fi
+
+    OPAMSWITCH=infer-$SUPP_OCAML_VERSION
+    opam switch install -j $NCPU $OPAMSWITCH --alias-of $ALIAS_SWITCH || true
```
jberdine: I have been testing this and have run into an issue.  The argument of --alias-of must be a compiler version, while ALIAS_SWITCH is the name of a switch, which is a general name and may not be a compiler version.  So for this functionality we need a method to produce a compiler version from a switch name.  Is this functionality that opam exposes, or is something more involved required?

jberdine:

build-infer.sh
jvillard: I found `opam config var ocaml-version --switch=infer-4.02.3`.

jvillard: Sorry, I misread; I think you want `opam config var ocaml-version`, and actually that's already computed in `$present` here. Does that fix the issue?

jberdine: Using `--alias-of $(opam config var ocaml-version)` is the same as the code before this PR.  I no longer understand how the original goal of this PR can be achieved.  Passing `--alias-of <version>` when `<version>` is already installed still fetches and builds ocaml in all scenarios I have tested.  @ronin13, what am I overlooking?

3flex:
I've tried looking at this, and it seems to be possible to figure out if the required version matches the version of the system compiler if it's acceptable that build-infer.sh would fail if a system compiler is not currently installed. But I don't think this is a great idea since it's easy to install opam without an installer system compiler.

What about adding an environment variable to build-infer.sh to control the behaviour? If set, can easily check the installed version and fail if it's not the supported version. This adds flexibility and the default would remain as-is.

jvillard:
Thanks for looking into this @3flex. Here's how I think about it, but I'm happy to change my mind if we find a solution that is more streamlined for everyone:
1. For default no-hassle installation, do `./build-infer.sh`
2. For all other worflows, there's an "expert mode", which is just the usual `./autogen.sh && ./configure && make && sudo make install` (`./autogen.sh` not needed for the releases). `./configure` will even suggest the right `opam` command to install the infer dependencies in the current switch if they are missing. It will also work for non-opam OCaml installations.

We could definitely advertise 2. more.

3flex:
Good to know - I was wondering if build-infer.sh was the canonical install method, but if there are two methods as you described that works.

I've been hacking on the Dockerfile a bit to try and bring the resulting image size down (it's currently 4GB+) and it would be easier to use the manual method to install in the Dockerfile than rely on build-infer.sh which is designed for ease of use. For that use case the manual method easily allows using the system compiler (bringing the image build time down as well).

I'll open a PR for the Dockerfile soon.

jvillard:
@ronin13  f4cbba109eec0c6cc1a352e7b7348bf4e3792bbd introduces two new options to `./build-infer.sh`, one of which lets you [set the opam switch](https://github.com/facebook/infer/commit/f4cbba109eec0c6cc1a352e7b7348bf4e3792bbd#diff-d58429b60862f9c8aeb07b70c8f37fb2R69) to whatever you like. You could use that to avoid rebuilding OCaml. Would that satisfy your use case?

3flex:
This works well for me, and I've updated #597 accordingly so OCaml is not rebuilt during the Docker image build.

The only deficiency is that it's not handled automatically, but I think given this is an "advanced" use case it's not a big loss.

cristianoc:
Is this still ongoing or should we close?

jvillard:
Closing as `./build-infer.sh` now accepts specifying a custom opam switch.

PR #472, opened 2016-09-30
Add assert macro into C model

master-q:
To use it as following example:

https://gist.github.com/master-q/d96b31d9e048a2c74abc70cdbfb3de46

### DIFF
--- a/infer/models/c/src/infer_builtins.h
+++ b/infer/models/c/src/infer_builtins.h
@@ -21,9 +21,10 @@ extern "C" {
 #endif
 
 // assert and if not, violate dereference pointer
-#define __infer_assert(E) \
-	int *s = NULL; \
-	if (!(E)) *s = 0xDEADBEEF;
+#define __infer_assert(E) {  \
+  int *s = NULL;             \
+  if (!(E)) *s = 0xDEADBEEF; \
+}
 
 // model returning an arbitrary (nondeterministic) short
 short __infer_nondet_short();
--- a/infer/models/c/src/infer_builtins.h
+++ b/infer/models/c/src/infer_builtins.h
@@ -20,6 +20,11 @@
 extern "C" {
 #endif
 
+// assert and if not, violate dereference pointer
+#define __infer_assert(E) \
+	int *s = NULL; \
+	if (!(E)) *s = 0xDEADBEEF;
+
 // model returning an arbitrary (nondeterministic) short
 short __infer_nondet_short();
 


jvillard:
Thanks for your PR @master-q 

I think distributing this with infer as-is may be problematic because assert failures are reported as null pointer dereferences, which is incorrect.

However, it looks like including this macro definition in your code as you're doing in your example produces the same result without having to modify infer at all. Does this cover your use-case well enough?

master-q:
> However, it looks like including this macro definition in your code as you're doing in your example
> produces the same result without having to modify infer at all. Does this cover your use-case well enough?

Yes. It's useful for me. Because the result shows the assertion failure location as following:

```
assert1.c:11: error: NULL_DEREFERENCE
  pointer s last assigned on line 11 could be null and is dereferenced at line 11, column 2
  9.    uid_t id;
  10.           id = getuid();
  11. >         __infer_assert(id != 0);
  12.   }
  13.   
```

Today, I'm using VeriFast https://people.cs.kuleuven.be/~bart.jacobs/verifast/ for my project. It can assert pre/postcondition. Of course, Infer also can do that. This patch's purpose  is to let beginers to be good with directions around such assertion on Infer.

jvillard:
Yes, I understand the value of having this macro in a project.

But I'm afraid it's still unclear to me what the value of hardcoding this in infer would be:
- it makes infer report the wrong kind of error (NPE instead of assertion failure), which is incorrect and confusing
-  unless I missed something, you still need to define that macro in your own code if you want it to compile, so isn't the situation pretty much the same with and without your patch?

master-q:
> it makes infer report the wrong kind of error (NPE instead of assertion failure), which is incorrect and confusing

Yes. It's possible.

> unless I missed something, you still need to define that macro in your own code if you want it to compile, so isn't the situation pretty much the same with and without your patch?

Yes. I can have it in my own project.
I'm afraid that many people believe Infer can't verify such assert, only do check memory error.

Totally, I think I should close this issue without merge, and write some blog entry at my web page.

http://metasepi.org/

Thanks a lot,

jvillard:
A blog post sounds great! Definitely this is something that is worth advertising.

master-q:
I write some precondition example:

https://github.com/master-q/practice-infer/blob/c642f4f92f80a4c2e19409fa8fef6be0d50c300f/DbC/main.c
https://github.com/master-q/practice-infer/blob/c642f4f92f80a4c2e19409fa8fef6be0d50c300f/DbC/queue.h

But Infer can't capture the violation. Why? Infer does not exhaustively verify invariant?

Issue #454, opened 2016-09-02
Support skip translation using path filters


kageiit:
Seems like the skip translation matchers only match with source file contents or method names

https://github.com/facebook/infer/blob/7e59032a427769e8003d8bf3c34d207b81f47d3d/infer/src/backend/inferconfig.ml#L196

This makes it very hard to exclude auto generated source files from codegen for skipping translation in Infer.

Seems like the a new `FilePatternMatcher` can be supported for this.

Once that is available, it would be nice to auto exclude generated source folders in build systems like buck automatically

jeremydubreil:
What kind of information could be used to distinguish those generated files ?

kageiit:
The best information to distinguish these files is the directory where they come from. All generated sources typically have an ancestor dir that contains them.

Looking at the generated source itself is not reliable because all annotation processors/gradle plugins etc do not produce similar generated code or annotate them the same way

jvillard:
Looks like this was fixed, there is now the `--skip-analysis-in-path` option.

PR #473, opened 2016-09-30
Add path filter

kageiit:
Ref #454

@jeremydubreil I got it close to a point where it can be finished up to merge. Im no ocaml expert and just based my code off the other existing ones.

Please let me know if something is completely off and what I should do to fix the last few compiler errors.

This will let us exclude entire paths like `buck-out/annotation` which contain all generated code and greatly improve infer run speeds

Thanks!


kageiit:
Closing since this was merged separately

Issue #36, opened 2015-06-12
Homebrew package


SeraphimSerapis:
It would be marvellous if Infer could be installed and updated via Homebrew.
Any plans to offer this as an alternative to compiling Infer yourself or retrieving the hosted binary?

a11n:
+1
...had the same thought after reading http://fbinfer.com/docs/getting-started.html

alistra:
:+1:

plowman:
Looks like someone is working on this here: https://github.com/Homebrew/homebrew/pull/40597

dambrisco:
@plowman As of right now, I'm basically blocked on building from source within Homebrew (see the previously linked issue for more context). I'm going to try to deep dive over the next week but debugging OPAM/OCaml and C is a tad bit outside of my wheelhouse. If you want to install the pre-built binary provided by Facebook, feel free to use my auxiliary tap: https://github.com/dambrisco/homebrew-aux

The existing PR into Homebrew is the same formula, but the maintainers have asked for a from-source build and I don't expect it'll be merged until I've managed to get that working.

tonycosentini:
I just opened https://github.com/Homebrew/homebrew/pull/47792 - it adds a formula that uses 0.5.

tonycosentini:
The Homebrew PR has landed.

You can now install Infer via Homebrew, just do:
``
brew update
brew install infer
```

If you want to just target Java you can do:
```
brew install infer --without-clang
```

or if you just want to target Clang you can do:
```
brew install infer --without-java
```

I think this can be closed now?

jvillard:
Thank you @tonycosentini :tada: :beer:

akotulski:
@tonycosentini - what do we do when we create new infer release?

tonycosentini:
If the build process does not change, we just need to change the download URL - https://github.com/Homebrew/homebrew/blob/1e0664f4b4ab73bb74b40e4f2d9370cc579e7986/Library/Formula/infer.rb

There are a few improvements that can be made to this script, I'm not a huge fan of listing the dependencies within the script since they are already in the OPAM file. Unfortunately, you can't init an OPAM repo in the same directory a package was pinned from.

tonycosentini:
For future reference, here's what the formula change looks like when there is a new release - https://github.com/Homebrew/homebrew/pull/48328

(I'll try to keep up with releases, just wanted to leave this for reference if anyone is interested.)

PR #490, opened 2016-10-27
Add how to resolve install error - "brew update"

ozmium:
If you get the error "Error: No available formula for infer . Searching taps... homebrew/science/infernal"
... Then you must run this command first:

brew update

And then run `brew install infer`. Reference: https://github.com/facebook/infer/issues/36


ozmium:
I have signed the Contributor License Agreement today.

PR #524, opened 2016-12-05
Update to Infer v0.9.4.1

bootstraponline:
Also the infer version seems to be reporting the wrong value.

```
$ infer -version
Infer version v0.9.4
```


jvillard:
Thanks @bootstraponline! Unfortunately, the version reported doesn't take into account the `.1` as you correctly point out...

PR #531, opened 2016-12-07
Clarify gradle command for Android Studio

ozmium:
The user must cd into the folder that contains 'gradlew' first, and then run the command, otherwise it doesn't work.
Also, the 2nd version of the command: "infer -- gradle build", does not seem to work for me. So move the working version to the top.


sblackshear:
Although you're right that you need to be in the same directory as the `gradlew` script in order to run it, I think I'd prefer to keep the shorter version of the documentation here.

Would you mind filing a separate issue describing what goes wrong with `infer -- gradle build` for you?

PR #535, opened 2016-12-10
Fix the widening operator of the Map functor

KihongHeo:
This commit avoids using the join operator for the widening 
of the Map functor in ```abstractDomain.ml```
and ensures termination when ```ValueDomain``` is infinite
by using ```ValueDomain.widen```.

### DIFF
--- a/infer/src/checkers/abstractDomain.ml
+++ b/infer/src/checkers/abstractDomain.ml
@@ -156,8 +156,17 @@ module Map (M : PrettyPrintable.PPMap) (ValueDomain : S) = struct
         astate1
         astate2
 
-  let widen ~prev ~next ~num_iters:_ =
-    join prev next
+  let widen ~prev ~next ~num_iters =
+    if prev == next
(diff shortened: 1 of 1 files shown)


sblackshear:
Thanks, good catch!

Could you please rebase on Infer master and use `phys_equal` instead of `==`? As of 7834c95bc831a41858e5088bb0795c8d2ebdb189 we no longer use `==`.

KihongHeo:
Hi @sblackshear ,
I rebased on Infer master and change ```==``` to ```phys_equal```. Please check it. 

By the way, I still cannot fully understand why do you use ```phys_equal``` instead of ```==```. 
Could you give me a reference for that?

jberdine:
`==` is syntactically very light-weight, and given that it has a tricky specification, should only be used in very specific cases.  Emphasizing the difference with `=` by using `phys_equal` instead is just a little mechanism to try to avoid incorrect use by making people thing twice before using it.

PR #549, opened 2016-12-27
Inferbo

KihongHeo:
This commit is for Inferbo: Infer-based buffer overrun analyzer.


infer/src/IR/Procdesc.re
```diff
@@ -519,3 +519,35 @@ let node_set_succs_exn pdesc (node: Node.t) succs exn =>
     set_succs_exn_base node' [exit_node] exn
   | _ => set_succs_exn_base node succs exn
   };
+
+let get_loop_heads pdesc => {
```
sblackshear: Will this work correctly for 

(1) Nested loops
(2) Loops with `break`/`continue`/`return`
(3) Loops inside of exceptional control-flow (e.g., `try { ... } catch (Exception _} { while (e1) { ... } }`)?

I ask (1)/(2) because the usual algorithm for computing loop heads in compilers is a bit more complicated than this; you need to compute the dominators of the graph first. I'm sure this simpler algorithm works in most cases, but for more complex loops I'm guessing you'll need the classic algorithm.

I ask (2) because I think `Node.get_succs` only returns the non-exceptional successors of the current node, which might cause the algorithm to miss loop heads inside of exceptional control-flow.

infer/src/IR/Procdesc.rei
```diff
@@ -166,6 +166,9 @@ let fold_calls: ('a => (Procname.t, Location.t) => 'a) => 'a => t => 'a;
 /** fold over all nodes and their instructions */
 let fold_instrs: ('a => Node.t => Sil.instr => 'a) => 'a => t => 'a;
 
+/** fold over all nodes */
+let fold_nodes: ('a => Node.t => 'a) => 'a => t => 'a;
```
sblackshear: Does this need to be exposed? I don't see any uses.

infer/src/backend/InferPrint.re
```diff
@@ -308,7 +308,8 @@ let should_report (issue_kind: Exceptions.err_kind) issue_type error_desc eclass
       switch Config.analyzer {
       | Checkers
       | Eradicate
-      | Tracing => true
+      | Tracing
+      | Bo => true
```
sblackshear: For consistency with the other (non-abbreviated) names, we probably want to call this something like `BufferOverrun` or `Buffers` or `Overrun`.

infer/src/bo/arrayBlk.ml
```diff
@@ -0,0 +1,210 @@
...
+  let top : t
+  = { offset = Itv.top; size = Itv.top; stride = Itv.top; }
+
+  let input : t
+  = { offset = Itv.zero; size = Itv.pos; stride = Itv.one; }
+
+  let make : Itv.t * Itv.t * Itv.t -> t
+  = fun (o, s, stride) -> { offset = o; size = s; stride = stride; }
+
+  let compare : t -> t -> int
```
sblackshear: You can avoid writing this `compare` function (and others) by adding the annotation `[@@deriving compare]` after the type declaration for `t` (see the Infer codebase for examples of this).

infer/src/base/Config.ml
```diff
@@ -497,6 +498,14 @@ and bootclasspath =
     ~exes:CLOpt.[Toplevel; Java]
     "Specify the Java bootclasspath"
 
+and bo_debug =
+  CLOpt.mk_int ~default:0 ~long:"bo-debug"
+    ~exes:CLOpt.[Toplevel] "Debug mode for buffer-overrun checker"
+
+and bo_filtering =
```
sblackshear: Infer has a builtin-in [mechanism](https://github.com/facebook/infer/blob/master/infer/src/IR/Localise.ml#L152) for filtering/prioritizing warnings. It would probably be best to re-use that mechanism rather than creating a new one just for buffer overruns.

infer/src/bo/basicDom.ml
```diff
@@ -0,0 +1,91 @@
+(*
```
sblackshear: A less generic filename would be useful here. Otherwise, it might be hard to tell which analyes should be using a `BasicDom`.

infer/src/bo/bufferOverrunChecker.ml
```diff
@@ -0,0 +1,470 @@
...
+    if Procname.get_method pname = "infer_print" then None else
+      match entry_mem, exit_mem with
+      | Some entry_mem, Some exit_mem ->
+          Summary.write_summary pname (entry_mem, exit_mem, cond_set);
+          Some (entry_mem, exit_mem, cond_set)
+      | _ -> None
+
+  let checker : Callbacks.proc_callback_args -> Dom.Summary.t option
+  = fun { get_proc_desc; proc_desc; proc_name; tenv } ->
+    let analyze_ondemand _ pdesc =
```
sblackshear: We recently made the `AbstractInterpreter` and `Interprocedural` module much more expressive so it should no longer be necessary to manually set up the on-demand callbacks like this. You should be able to add all of the postprocessing steps like reporting in a custom `compute_post` function that you pass via the `~compute_post` parameter to `Interprocedural`. See siof.ml and taintAnalysis.ml for examples of this.

infer/src/checkers/AbstractDomain.ml
```diff
@@ -87,6 +87,38 @@ module Pair (Domain1 : S) (Domain2 : S) = struct
     F.fprintf fmt "(%a, %a)" Domain1.pp astate1 Domain2.pp astate2
 end
 
+module Pair3 (Domain1 : S) (Domain2 : S) (Domain3 : S) = struct
```
sblackshear: Wondering if we need this. Two invocations of the `Pair` functor can achieve the same thing, and for domains with > 2 components I think it's typically cleaner to use a record type for `astate` rather than a tuple. The advantage of using a record is that it's much easier to extend than a pair, and it's easier to ignore parts of the domain you don't care about without writing a bunch of `_`'s.

infer/tests/codetoanalyze/c/bo/out
```diff
@@ -0,0 +1,9307 @@
+Capturing in make/cc mode...
```
sblackshear: Are these files meant to be checked in? What are they for?

infer/tests/codetoanalyze/c/bo/test2.c
```diff
@@ -0,0 +1,18 @@
...
+
+int bar(int argc, char**argv){
+  int arr[10];
+  arr[0] = 100;
+  char* p = malloc(10);
+
+  foo(arr, p, 20);
+
+  return 0;
+}
```
sblackshear: Adding some more tests would be really helpful for ensuring that we don't break existing functionality of inferbo. A few that I'd like to see

- Really simple tests reading directly from a bad index with no loop, like `int arr[1]; arr[1] = 2`
- Do..while loops
- Nested loops
- Loops with `break`/`continue`/`return`
- Nested loops with labeled `break`/`continue` (e.g., continue to outer loop head from inner loop)
- Infinite loops
- Unstructured loops formed using `goto`


It's ok if the analysis can't handle all of these features precisely, but it's really useful to have tests that make the capabilities and limitations of the analysis clear without actually reading the code for the analysis. Tests demonstrating a limitation are nice for pointing someone looking to get involved to a concrete improvement they can make.

infer/src/bo/arrayBlk.ml
KihongHeo: In this case [https://github.com/ropas/infer/blob/inferbo/infer/src/bo/itv.ml#L830](url),
how can I use the annotation ```[@@deriving compare]```? 
The following code 
```ocaml
type t = astate [@@deriving compare]
```
yields 
```Error: Unbound value compare_astate```

File: infer/src/IR/Procdesc.re (hunk 1 above)
Comment (skcho) [inline infer/src/IR/Procdesc.re]: (1) Yes.
(2) Yes if `Node.get_succs` returns edges of `break` / `continue` / `return`.

The classical algorithm for computing loop heads in compilers is not good for setting widening points because it does not catch all cycles. For example,
```c
if(cond){
a:
  goto b;
}else{
b:
  goto a;
}
```
there is a cycle between `a` and `b`, so one of them should be set as a widening point. However, they cannot be loop heads by the classical algorithm because `a` and `b` do not dominate each other.

While the classical algorithm finds loop heads by back edges to dominators, our algorithm finds all nodes that have back edges when drawing depth-first spanning tree.

(3) You are right. We added `Node.get_exn` to find exceptional control-flow.

infer/src/bo/bufferOverrunChecker.ml
skcho: We changed to use the `AbstractInterpreter.Interprocedural` functor. Thanks.

jberdine:
> While the classical algorithm finds loop heads by back edges to dominators, our algorithm finds all nodes that have back edges when drawing depth-first spanning tree.

FWIW, selecting the targets of back-edges in a depth-first traversal is the algorithm I settled on in the past.

infer/tests/codetoanalyze/c/bufferoverrun/Makefile
```diff
@@ -0,0 +1,29 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+TESTS_DIR = ../../..
+
+ANALYZER = bufferoverrun
+# see explanations in cpp/errors/Makefile for the custom isystem
```
akotulski: this comment doesn't make sense - it's not needed in this case

infer/tests/codetoanalyze/c/bufferoverrun/main.c
```diff
@@ -0,0 +1,23 @@
...
+void for_loop();
+void while_loop();
+void do_while();
+void nested_loop();
+void nested_loop_with_label();
+int break_continue_return();
+void goto_loop();
+void inf_loop();
+
+int main() {
```
akotulski: why do you need this, doesn't analyzer work without `main`? What happens if you rename `main` to `run`?

infer/tests/codetoanalyze/c/bufferoverrun/break_continue_return.c
```diff
@@ -0,0 +1,18 @@
+int break_continue_return(){
```
akotulski: please run`clang-format -i <FILE>` on all c files. You can use one from `path-to-infer/facebook-clang-plugins/clang/install/bin`

It will probably change `.exp` file so remember to change it

infer/src/Makefile
```diff
@@ -123,7 +123,7 @@ else
 EXTRA_DEPS = opensource
 endif
 
-DEPENDENCIES = IR backend base checkers eradicate harness integration tp/fts quandary $(EXTRA_DEPS)
+DEPENDENCIES = IR backend base checkers eradicate harness integration tp/fts quandary bo $(EXTRA_DEPS)
```
akotulski: minor: should we have directory with more self explanatory name?

infer/tests/codetoanalyze/c/bufferoverrun/main.c
skcho: No, we does not need the `main` function indeed. We removed it.

infer/src/Makefile
skcho: We renamed it to `bufferoverrun`.

infer/src/IR/Procdesc.re
sblackshear: Ok, thanks. That makes sense. I should have paid more attention to what the code is doing.

Thanks for contributing this! Widening only at loop heads is a huge improvement over what we do today (widening every time a node is visited >1 time).

infer/src/base/Config.ml
KihongHeo: Could you give me a pointer that uses Bucket to filter out alarms?

sblackshear: The two best files to look at are localise.ml (see the `Buckets` module) and errdesc.ml (see the `explain` functions and their usages from other modules. 

Although things are a bit disorganized, the typical pattern is to assign a warning to a bucket from B1 (highest confidence) to B5 (highest confidence) while creating a `Localise.error_desc`. The actual filtering (which shouldn't need to be changed) happens in `Exceptions.recognize_exception`. By default, buckets B1 and B2 are reported, and higher buckets are filtering unless the `--no-filtering` flag is passed.

KihongHeo: ```Config.filter_buckets``` in ```Exceptions.recognize_exception``` is always false so nothing is filtered in B3~B5. Could you check it?

sblackshear: Oops, I pointed you to the wrong place. You are right that this code is useless; good catch. I meant to point you to [here](https://github.com/facebook/infer/blob/master/infer/src/backend/InferPrint.re#L344).

KihongHeo: Thanks. I change the filtering the filtering by using buckets. 
I think we have addressed all of the issues in this thread. Do you have other comments?

infer/src/bufferoverrun/bufferOverrunDomain.ml
```diff
@@ -111,17 +111,19 @@ struct
...
+    then None
+    else if filter1 c then Some 0
+    else if filter2 c then Some 1
     else
       let c = set_size_pos c in
       let not_overrun = Itv.lt_sem c.idx c.size in
       let not_underrun = Itv.le_sem Itv.zero c.idx in
-      (Itv.eq not_overrun Itv.one) && (Itv.eq not_underrun Itv.one)
+      if (Itv.eq not_overrun Itv.one) && (Itv.eq not_underrun Itv.one) then None
+      else Some 2
```
sblackshear: Perhaps this should be a new variant type if you are only using values from 1-2? Or I suppose you could use the bucket values directly (assuming that doesn't introduce circular dependencies).

sblackshear:
Nope, I don't think there's much more to address! Just the last small comment and a rebase on Infer master, then I will import the pull request and run some some internal tests.

KihongHeo:
I appreciate your help. I addressed your last comment and rebased it on Infer master. Thanks.

jvillard:
It looks like GitHub thinks you added commits to your branch instead of rebasing. Could you try again? You should be able to rebase your branch on top of infer/master, then force push it back to GitHub.

KihongHeo:
I cannot understand what you mean. (Sorry, I am not that familiar with GitHub.) The current branch is  rebased on top of infer/master. I can see the latest commit of infer/master with ```git log``` in my branch. Could you give me some advises?

jvillard:
If you look at the commits in your branch, there are also commits from infer/master, which shouldn't be the case: https://github.com/facebook/infer/pull/549/commits

To rebase on top of infer:master from within your repo, something like this should work:

```
$ # if you haven't already:
$ git remote add fbinfer https://github.com/facebook/infer
$ git fetch fbinfer
$ git checkout inferbo # go to your branch
$ git rebase fbinfer/master # solve conflicts, etc.
$ git push -f # push to https://github.com/ropas/infer
```

KihongHeo:
I think it is fixed. There are no more Infer team's commits (https://github.com/facebook/infer/pull/549/commits). Could you check it again?

sblackshear:
Just an update: everything looks good on your end, but we're still running some internal tests.

PR #553, opened 2017-01-03
Add bsdtar to dependency list

napfi:
Infer is packed with BSD tar, which makes to GNU tar emit error messages about unknown extended header keywords.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -44,6 +44,7 @@ is required to be able to [use the
 release](http://fbinfer.com/docs/getting-started.html) (faster), or to
 compile everything from source (see the end of this document).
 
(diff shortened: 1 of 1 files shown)


jvillard:
Thanks for the report. I agree that these messages are annoying. I'll reupload the archive using GNU tar.

Regarding the PR, I'm not sure BSD tar qualifies as a dependency, as infer works without it. In fact, you can even get rid of the warnings with `--warning=no-unknown-keyword` in GNU tar.

jvillard:
Offending tarball has been re-uploaded with GNU tar and the annoying messages are gone, so I don't think this PR is necessary anymore.

PR #554, opened 2017-01-05
Update build-infer.sh

chaws:
The guidelines say I make sure infer builds, but this PR is exactly to fix a tiny issue with the build-infer.sh script. Changes are numbered below:

1. **opam init**: '--compiler' option has been changed to '--comp' (ref: https://github.com/ocaml/opam/commit/36a5d5f5a68d29b882bd4a584f66c0ae18235c6c)
   
2. **opam switch**: was giving 'too many arguments' error. The 'set' submenu is optional now, so I just removed that and rearranged arguments order

### DIFF
--- a/build-infer.sh
+++ b/build-infer.sh
@@ -98,10 +98,10 @@ check_installed () {
 setup_opam () {
     OCAML_VERSION="4.02.3"
 
-    opam init --compiler=$OCAML_VERSION -j $NCPU --no-setup --yes
+    opam init --comp=$OCAML_VERSION -j $NCPU --no-setup --yes
 
     OPAMSWITCH=infer-$OCAML_VERSION
-    opam switch set -j $NCPU $OPAMSWITCH --alias-of $OCAML_VERSION
+    opam switch $OPAMSWITCH -j $NCPU --alias-of=$OCAML_VERSION
 }
 
 add_opam_git_pin () {


jvillard:
Thanks! Which version of `opam` are these changes needed for?

chaws:
I'm actually not sure...I'm running `1.2.0` (as required by infer), and found out that commit on their end (https://github.com/ocaml/opam/commit/36a5d5f5a68d29b882bd4a584f66c0ae18235c6c) that seems to have been done to go with `1.2.1~beta`.

jvillard:
I see. I wonder if it would be better to bump the required version of opam instead of trying to support older versions like 1.2.0. I use 1.2.2 and I don't see this issue. Otherwise we might break opam 1.2.0 again in the future without noticing it.

jvillard:
@chaws does upgrading opam solve the issue for you? If so, could you please change your PR to bump the required version of opam up?

As it stands, opam 1.2.0 is untested on our end so even if we include your fix it's possible we end up breaking it again in the future. 1.2.2 is available on many major systems. Debian stable still uses 1.2.0 but it's easy to manually upgrade.

Let me know if that makes sense to you, we could also find a way to test opam 1.2.0 in our Travis CI for instance.

chaws:
Hi @jvillard , I didn't have much time over the weekend, Let me update my opam (I'll install from the source) and give you a feedback.

chaws:
@jvillard , how did you get opam 1.3?
I'm trying to build it, but keep getting:
File "solver/opamSolver.ml", line 371, characters 10-12:
Error: The variant type Algo.Diagnostic.request has no constructor ::
Any insights?

chaws:
@jvillard , I'm trying to get opam 1.3.0 out of the source () but keep getting this error:
```
File "solver/opamSolver.ml", line 371, characters 10-12:
Error: This variant pattern is expected to have type Diagnostic.request
       The constructor :: does not belong to type Algo.Diagnostic.request
../OCamlMakefile:1076: recipe for target 'solver/opamSolver.cmo' failed
```

Which seems like a syntax issue. Ocaml is new to me, I'm trying to grasp up some of it thru tutorials, but if you already have a hunch on solving this, that'd be great!

jvillard:
I don't see any 1.3.0 release [here](http://opam.ocaml.org/doc/Install.html#FromSources). I think the most recent one is 1.2.2.

There are also [binary distributions](http://opam.ocaml.org/doc/Install.html#Binarydistribution) to avoid having to compile from source.

chaws:
Installing 1.2.2 from [here](http://opam.ocaml.org/doc/Install.html#FromSources) seemed to fix those flags!

Thanks!

PR #557, opened 2017-01-06
Add an option to skip updating submodules if only building for java

kageiit:
When invoking `./build-infer.sh java`, there is no need to update the facebook-clang-plugins submodule that is done as part of the `autogen.sh` script. The submodule update takes a very long time and can be skipped when building only for java


kageiit:
@sblackshear Can this be merged?

PR #558, opened 2017-01-06
Cleanup DEFS file for generating infer genrules

kageiit:
- Only generate one extra genrule for running infer. Remove all other java library rules currently being generated
- Generate infer genrule only if the `java_library` has `srcs`, otherwise there is nothing to analyze
- Use `SRCDIR` to avoid making a copy of the target sources as buck will just symlink them instead
- Added support for `android_library` rules as well
- Added support to generate both `infer` and `eradicate` genrules

### DIFF
--- a/DEFS
+++ b/DEFS
@@ -1,5 +1,7 @@
 import os
 
+analyzers = ['infer', 'eradicate']
+
 original_java_library = java_library
 def java_library(
     name,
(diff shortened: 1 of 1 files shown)


kageiit:
Will make another change for the ability to add `.inferconfig` and `.inferversion` etc to the `srcs` as that will cause the genrule to invalidate when the infer version or configuration changes

PR #581, opened 2017-02-13
Excluding delegates storages from STRONG_DELEGATE_CHECK

RavisMsk:
Hi!
It's quite common to have collections of delegates. The collection itself is usually named like  "_delegates_Hash", "_delegates_Storage" or simply "_delegates_". Obviously, there is common part in all these cases, but currently you're excluding property only if it contains "queue".
I've added a simple exclusion by common part. It solved our warnings and I think for others it'll be quite helpful too.

P.S. I made that pull request to 0.9.5 since it is currently released version and master branch has separate rules file. I can make separate pull request for master with the same simple fix.

### DIFF
--- a/infer/src/clang/cFrontend_checkers.ml
+++ b/infer/src/clang/cFrontend_checkers.ml
@@ -166,13 +166,16 @@ let ctl_strong_delegate_warning lctx an =
   let open CTL in
   let name_contains_delegate =
     Atomic ("property_name_contains_word", ["delegate"]) in
+  let name_does_not_contain_delegates =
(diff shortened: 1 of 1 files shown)


dulmarod:
Hi, thanks for your pull request. Unfortunately, this check has changed in master. We are working on a DSL for writing linters, and in the latest version this check has already been written in the DSL. If you are interested, you could try to download and build master, and make the pull request on top of it. Thanks.

RavisMsk:
@dulmarod

dulmarod:
By the way, in master the check leaves in infer/lib/linter_rules/linters.al.

PR #583, opened 2017-02-14
Add SIOF explanation to checkers docs

akotulski:


akotulski:
I'd appreciate if someone looked over this change - it's very hard to explain what SIOF is...

PR #585, opened 2017-02-23
Docker: Update to Infer v0.10.0

3flex:
Given #524 was required to do the last manual update, perhaps the Dockerfile could be updated as part of the release process instead of catching it after the fact?

Cheers!

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -44,7 +44,7 @@ RUN VERSION=1.2.2; \
 RUN opam init -y --comp=4.02.3
 
 # Download the latest Infer release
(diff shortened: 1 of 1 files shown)


jvillard:
Thanks @3flex, I was updating these files as well but you submitted the PR before I got the chance to land my (identical) changes. I also wanted to make sure that Travis gave better signal with 1e14c87a729bb65cf4d23d7f7fc46246b049e180.

There are a few steps to coordinate when making a release (GitHub, Homebrew, and Docker, for now), and they cannot happen atomically all at once (GitHub needs to happen before Homebrew and Docker).

I'll merge in your changes.

PR #589, opened 2017-02-24
Update supported opam version to 1.2.2

### LINKED ISSUE #588: build-infer.sh fails with opam 1.2.0
Related to #554 (specifically this https://github.com/facebook/infer/pull/554#issuecomment-271283944)

Confirm that build-infer.sh fails on opam 1.2.0 due to the changed name of the flag as of https://github.com/ocaml/opam/commit/9ccabc03f75b1908ce5fda4c676ddd16d9d0c7ce#diff-906c973c07f8c64e8dd7866de2b40e35. Error is:
```
initializing opam... 
opam: unknown option `--compiler'
```

Would either need to support both `--comp` and `--compiler` flags in build-infer.sh so Debian stable continues to work, or increase minimum opam version to 1.2.1 in [INSTALL.md](https://github.com/facebook/infer/blob/master/INSTALL.md)

3flex:
Fixes #588

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,7 +18,7 @@ is required to be able to [use the
 release](http://fbinfer.com/docs/getting-started.html) (faster), or to
 compile everything from source (see the end of this document).
 
-- opam >= 1.2.0 (instructions [here](https://opam.ocaml.org/doc/Install.html#OSX))
+- opam 1.2.2 (instructions [here](https://opam.ocaml.org/doc/Install.html#OSX))
 - Python 2.7
 - pkg-config
 - libffi >= 3.0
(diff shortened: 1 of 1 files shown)


jvillard:
As discussed in #588, can you update this so that it bumps the minimum required version in the instructions instead? Thanks!

3flex:
Version 1.2.1 or 1.2.2? Not really sure what the difference is but either one would fix at least the `--comp`/`--compiler` discrepancy

3flex:
Oh and also with opam 2 in beta, and presumably untested by Infer package process, 1.2.x would the only supported version for now?

3flex:
Updated, assuming 1.2.2 should be the only supported version for now.

jvillard:
Looks good, thanks!

PR #595, opened 2017-02-28
Update Android Plugin for Gradle (APG) and dependencies

3flex:
* Makes running Android example simpler - APG 2.2.0 automatically downloads dependencies so long as the user has previously accepted the Android SDK License. 
* APG 2.2.0 requires Gradle 2.14 or newer, so updated to latest.
* Added `buildOptions.abortOnError false` so the build doesn't report any lint issues - the sample is designed to show Infer working, not Android lint.

### DIFF
--- a/examples/android_hello/app/build.gradle
+++ b/examples/android_hello/app/build.gradle
@@ -3,6 +3,8 @@ apply plugin: 'com.android.application'
 android {
     compileSdkVersion 22
     buildToolsVersion "22.0.1"
+    
+    buildOptions.abortOnError false
 
     defaultConfig {
         applicationId "infer.inferandroidexample"
--- a/examples/android_hello/build.gradle
+++ b/examples/android_hello/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
+        classpath 'com.android.tools.build:gradle:2.2.3'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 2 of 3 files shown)


jvillard:
Looks good but Travis failed and doesn't show me why (https://travis-ci.org/facebook/infer/builds/206290395 gives me "There was an error while loading data.").

> Makes running Android example simpler - APG 2.2.0 automatically downloads dependencies so long as the user has previously accepted the Android SDK License.

Could you elaborate in what way is running the example simpler?

3flex:
Doesn't look like Travis runs the Android test, so can't see how this would fail the build - it may be due to the AWS outage that started around when I submitted the PR (https://www.traviscistatus.com/incidents/hmwq9yy5dh9d) - can you rerun the failing job?

It's simpler because the user doesn't need to manually install build tools 22.0.1, SDK 22, or the Android Support Repository - they'll be automatically downloaded on `./gradlew build`, as long as the user has previously agreed to the SDK license on the machine they're using. It will "just work".

The main benefit for me is that this allows for simplification of the Docker image; the dependencies can be downloaded automatically if the user runs the Android sample, which means they can be kept out of the image to keep the size of the image down, though granted it will be slower to run. I'm about to submit a PR for some proposed Dockerfile changes.

jvillard:
Thanks for the explanations @3flex. I don't see a way for me to rerun the test, I'll just test it locally.

3flex:
@jvillard this is the faulty PR which says `buildOptions` not `lintOptions`, but until it's merged to master I'm not sure the best way to open a new PR to correct it?

There might be a way but my git knowledge only goes so far :/

jvillard:
My bad, I thought this had been merged already. I can amend it to use `lintOptions` and push it.

PR #597, opened 2017-03-01
Dockerfile update

3flex:
This cuts the resulting image from ~4.6GB to ~1.6GB.

Infer functionality is not sacrificed and the following commands run successfully without any extra user interaction (including Android example provided the user has accepted the Android SDK license which they'll be prompted to do when running the Docker image, see https://github.com/facebook/infer/pull/597#issuecomment-284864016):
```bash
cd /infer/examples && infer -- javac Hello.java
cd /infer/examples && infer -- gcc -c hello.c
cd /infer/examples && infer -- clang -c hello.c
cd /infer/examples/c_hello && infer -- make
cd /infer/examples/java_hello && infer -- javac Hello.java Pointers.java Resources.java
cd /infer/examples/android_hello && infer -- ./gradlew build
```

The OPAM installation is no longer retained. This may impact Infer developers if they're using the Docker images and require OPAM packages to be installed as part of their workflow.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,6 +6,7 @@ MAINTAINER Infer
 # Debian config
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+            aspcud \
             autoconf \
             automake \
             gcc \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -44,7 +44,8 @@ RUN INFER_VERSION=v0.10.0; \
 
 # Compile Infer
 RUN cd /infer && \
-    ./build-infer.sh
+    ./build-infer.sh && \
+    rm -rf /root/.opam
 
 # Install Infer
 ENV INFER_HOME /infer/infer
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,11 +8,10 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
             autoconf \
             automake \
-            curl \
             gcc \
             g++ \
             git \
-            groff \
+            groff-base \
             libc6-dev \
             libffi-dev \
             libgmp-dev \
@@ -25,11 +24,11 @@ RUN apt-get update && \
             opam \
             openjdk-8-jdk-headless \
             pkg-config \
-            python-software-properties \
+            python2.7 \
             rsync \
-            software-properties-common \
             unzip \
-            zlib1g-dev
+            zlib1g-dev && \
+    rm -rf /var/lib/apt/lists/*
 
 RUN opam init -y --comp=4.02.3
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,8 +40,10 @@ RUN INFER_VERSION=v0.11.0; \
     ln -s ${PWD}/infer-linux64-$INFER_VERSION /infer
 
 # Compile Infer
-RUN cd /infer && \
-    ./build-infer.sh && \
+RUN OCAML_VERSION=4.04.0; \
+    opam init --compiler=$OCAML_VERSION -j $(getconf _NPROCESSORS_ONLN || echo 1) --yes && \
+    cd /infer && \
+    ./build-infer.sh --opam-switch $OCAML_VERSION && \
     rm -rf /root/.opam
 
 # Install Infer
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,6 +22,7 @@ RUN apt-get update && \
             make \
             ncurses-dev \
             ocaml \
+            opam \
             openjdk-8-jdk-headless \
             pkg-config \
             python-software-properties \
@@ -30,14 +31,6 @@ RUN apt-get update && \
             unzip \
             zlib1g-dev
 
-# Install OPAM
-RUN VERSION=1.2.2; \
-    curl -sL \
-      https://github.com/ocaml/opam/releases/download/$VERSION/opam-$VERSION-x86_64-Linux \
-      -o /usr/local/bin/opam && \
-    chmod 755 /usr/local/bin/opam && \
-    ((/usr/local/bin/opam --version | grep -q $VERSION) || \
-     (echo "failed to download opam from GitHub."; exit 1))
 RUN opam init -y --comp=4.02.3
 
 # Download the latest Infer release
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,18 +1,13 @@
 # Base image
-FROM heikomaass/android-sdk
+FROM buildpack-deps:xenial-curl
 
 MAINTAINER Infer
-
-# Add android-22 and build-tools-22 to the Android SDK
-RUN ["/opt/sdk-tools/android-accept-licenses.sh", \
-     "android update sdk --filter \"android-22\" --no-ui --force --all"]
-RUN ["/opt/sdk-tools/android-accept-licenses.sh", \
-     "android update sdk --filter \"build-tools-22.0.1\" --no-ui --force --all"]
 
 # Debian config
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
             autoconf \
+            automake \
             curl \
             gcc \
             g++ \
@@ -27,6 +22,7 @@ RUN apt-get update && \
             make \
             ncurses-dev \
             ocaml \
+            openjdk-8-jdk-headless \
             pkg-config \
             python-software-properties \
             rsync \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -55,8 +41,23 @@ RUN INFER_VERSION=v0.10.0; \
...
+    apt-get install -y --no-install-recommends \
+            libc6-dev \
+            libc6-i386 \
+            lib32z1 \
+            lib32stdc++6 && \
+    rm -rf /var/lib/apt/lists/*
+
+ENV ANDROID_HOME /opt/android-sdk-linux
+RUN mkdir -p /opt/android-sdk-linux/licenses
+RUN echo -e "\n8933bad161af4178b1185d1a37fbf41ea5269c55" > ${ANDROID_HOME}/licenses/android-sdk-license
```
jvillard: How does one get this number?

jvillard:
I get an error with this change:

```sh
$ cd examples/android_hello/
$ ./gradlew clean

FAILURE: Build failed with an exception.

* Where:
Build file '/home/jul/infer/examples/android_hello/app/build.gradle' line: 7

* What went wrong:
A problem occurred evaluating project ':app'.
> Could not get unknown property 'buildOptions' for object of type com.android.build.gradle.AppExtension.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 0.704 secs
$ ./gradlew --version

------------------------------------------------------------
Gradle 3.4
------------------------------------------------------------

Build time:   2017-02-20 14:49:26 UTC
Revision:     73f32d68824582945f5ac1810600e8d87794c3d4

Groovy:       2.4.7
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          1.8.0_121 (Oracle Corporation 25.121-b13)
OS:           Linux 4.9.0-1-amd64 amd64
```

3flex:
Ah I'm sorry, that error is because of human error in #595... it should have been `lintOptions.abortOnError false` not `buildOptions.abortOnError false`.

I can open a PR to fix, you can revert #595 and I'll resubmit that PR, or I can fix it here?

Regarding the magic number for the license, that's the hash of the license text for the current Android SDK License. Normally user agrees in GUI in Android Studio, or via `sdkmanager` CLI, then `licenses/android-sdk-license` is populated with that hash. In this script it's just being done automatically, replacing `/opt/sdk-tools/android-accept-licenses.sh` in the current image. See https://developer.android.com/studio/intro/update.html#download-with-gradle.

This is actually shaky ground because the user never explicitly agrees to the Android license (the current Dockerfile has this problem too, the license is automatically agreed to without the user actually doing it). So you may want to remove this functionality of agreeing to the license automatically?

One other option (I'm checking now) is installing the Android dependencies from the Ubuntu package repo instead for purposes of having a working sample in the image, it might avoid the licensing problem.

jvillard:
Thanks @3flex.

Please open a new PR to fix the gradle build.

Regarding the magic hash, maybe there could be a build argument that auto-accepts the license when passed. However, is it really needed? It looks like the user could accept the license only at the time when they run the Android example, at which point they have an interactive shell to accept it.

Probably the interactive step doesn't work inside `infer -- ./gradlew` because infer doesn't pass stdin to the build command, so `./run.sh` has to be modified to have a step before the analysis calling gradle in a minimal way just to give the user a chance to accept the license (if that makes sense).

3flex:
I have a solution I'm happy with, but it depends on Android SDK Tools 25.3.x which has not yet been officially released.

Also depends on a new release of Infer with the Android Plugin for Gradle updates I'd submitted included so `infer -- ./gradlew build` will just work.

Once those are both out I'll update this PR.

What it does is show a message after build completes in `run.sh`, so you get a prompt like:
```
Successfully built 2c179014c138
*************************************************************
To build the Android example, you must accept the Android SDK
licenses by running 'sdkmanager --licenses' first.
*************************************************************
root@ecb387440a1f:/infer/examples# 
```

This relies on functionality in `sdkmanager` that was introduced in the new SDK Tools, allowing for easily accepting all licenses in one hit. I think it neatly solves the problem.

It means another ~140MB in the image so the SDK Tools can be preinstalled but it's still a big improvement over what's currently there.

3flex:
@jvillard are there plans to cut a new Infer release? I'll update the PR as soon as that happens.

jvillard:
@3flex FYI I just released 0.11.0.

3flex:
Is OCaml 4.04.0 the minimum supported now? I'll have to make changes if so.

I couldn't tell from release notes if it's just compatible with 4.04.0 or whether that's the minimum required version.

jvillard:
@3flex We pretty much only support one OCaml version at a time and now that version is 4.04.0 :)

3flex:
ping @jvillard :)

jvillard:
Looks awesome! Thanks a lot, I'm trying it now.

jvillard:
I see this error while it's compiling infer:
```
[*ERROR**][18315] ocamlfind ocamlopt -c -g -short-paths -safe-string -principal -strict-formats -strict-sequence -w +3+5+6+8+10+11+12+18+19+20+23+26+29+27+33+34+35+37+38+39+50-4-9-32-40-41-42-45-48 -bin-annot -thread -package sawja -package ptrees -package javalib -package zip -package yojson -package xmlm -package unix -package str -package oUnit -package extlib -package ctypes.stubs -package core -package atdgen -package ANSITerminal -package ppx_compare -pp '-g -short-paths -safe-string -principal -strict-formats -strict-sequence -w +3+5+6+8+10+11+12+18+19+20+23+26+29+27+33+34+35+37+38+39+50-4-9-32-40-41-42-45-48 -bin-annot -thread -package sawja -package ptrees -package javalib -package zip -package yojson -package xmlm -package unix -package str -package oUnit -package extlib -package ctypes.stubs -package core -package atdgen -package ANSITerminal -package ppx_compare' -I clang -I ffi -I quandary -I base -I harness -I integration -I IR -I eradicate -I opensource -I checkers -I backend -I bufferoverrun -I java -I ffi/gen -I unit/clang -pp 'refmt --print binary' -o clang/ClangCommand.cmx -intf-suffix .rei -impl clang/ClangCommand.re
[*ERROR**][18315] ocamlfind ocamlopt -c -g -short-paths -safe-string -principal -strict-formats -strict-sequence -w +3+5+6+8+10+11+12+18+19+20+23+26+29+27+33+34+35+37+38+39+50-4-9-32-40-41-42-45-48 -bin-annot -thread -package sawja -package ptrees -package javalib -package zip -package yojson -package xmlm -package unix -package str -package oUnit -package extlib -package ctypes.stubs -package core -package atdgen -package ANSITerminal -package ppx_compare -w -27-32-34-35-39 -I clang -I ffi -I quandary -I base -I harness -I integration -I IR -I eradicate -I opensource -I checkers -I backend -I bufferoverrun -I java -I ffi/gen -I unit/clang -o clang/clang_ast_b.cmx clang/clang_ast_b.ml
[*ERROR**][18315] ocamlfind ocamlopt -c -g -short-paths -safe-string -principal -strict-formats -strict-sequence -w +3+5+6+8+10+11+12+18+19+20+23+26+29+27+33+34+35+37+38+39+50-4-9-32-40-41-42-45-48 -bin-annot -thread -package sawja -package ptrees -package javalib -package zip -package yojson -package xmlm -package unix -package str -package oUnit -package extlib -package ctypes.stubs -package core -package atdgen -package ANSITerminal -package ppx_compare -I backend -I ffi -I quandary -I base -I harness -I integration -I IR -I eradicate -I opensource -I checkers -I bufferoverrun -I clang -I java -I ffi/gen -I unit/clang -o backend/PerfStats.cmx backend/PerfStats.ml
[*ERROR**][18315] + ocamlfind ocamlopt -c -g -short-paths -safe-string -principal -strict-formats -strict-sequence -w +3+5+6+8+10+11+12+18+19+20+23+26+29+27+33+34+35+37+38+39+50-4-9-32-40-41-42-45-48 -bin-annot -thread -package sawja -package ptrees -package javalib -package zip -package yojson -package xmlm -package unix -package str -package oUnit -package extlib -package ctypes.stubs -package core -package atdgen -package ANSITerminal -package ppx_compare -w -27-32-34-35-39 -I clang -I ffi -I quandary -I base -I harness -I integration -I IR -I eradicate -I opensource -I checkers -I backend -I bufferoverrun -I java -I ffi/gen -I unit/clang -o clang/clang_ast_b.cmx clang/clang_ast_b.ml
[*ERROR**][18315] Fatal error: exception Invalid_argument("output_value: abstract value (outside heap)")
[*ERROR**][18315] Command exited with code 2.
[*ERROR**][18315] Makefile:227: recipe for target '/opt/infer-linux64-v0.11.0/infer/_build/infer/backend/infer.native' failed
[*ERROR**][18315] make[1]: Leaving directory '/opt/infer-linux64-v0.11.0/infer/src'
[...]
The command '/bin/sh -c cd /infer &&     ./build-infer.sh &&     rm -rf /root/.opam' returned a non-zero code: 1
```

Retrying now.

docker/Dockerfile
```diff
@@ -55,8 +41,27 @@ RUN INFER_VERSION=v0.10.0; \
 
 # Compile Infer
 RUN cd /infer && \
-    ./build-infer.sh
+    ./build-infer.sh && \
```
jvillard: I think this can be a bit more efficient because here we're compiling OCaml twice. Instead, you can set up the switch in advance and then tell `./build-infer.sh` that that switch has already been configured and is to be used to build infer:
```sh
RUN OCAML_VERSION=4.04.0; \
    opam init --compiler=$OCAML_VERSION -j $(getconf _NPROCESSORS_ONLN || echo 1) --yes && \
    cd /infer && \
    ./build-infer.sh --opam-switch $OCAML_VERSION && \
    rm -rf /root/.opam
```

(We don't do this by default to ensure the infer install is somewhat sandboxed, although I imagine we could do something like that when we detect that opam is not set up at all...)

jvillard:
I'm noticing that camlp4 often fails to compile at first, then succeeds the second time: https://gist.github.com/jvillard/3aa6397f76fed1e3275c0ea484a20739

I don't think it's related to your change though, just making a note for future reference...

3flex:
Updated as suggested - OCaml now only builds once, thanks for the recommendation!

I didn't see the issue building camlp4... strange it only occurs sometimes.

jvillard:
Thanks a lot for your contribution @3flex, that's a significant improvement (and an upgrade to 0.11.0 at the same time!).

PR #599, opened 2017-03-02
Infer is C++ now

jvillard:


_config.yml
```diff
@@ -1,9 +1,9 @@
 # Site settings
 permalink: /blog/:year/:month/:day/:title.html
 title: Infer
-tagline: A tool to detect bugs in Android and iOS apps before they ship
+tagline: A tool to detect bugs in Java and C/C++/Objective-C apps before they ship
```
akotulski: maybe not use `apps` anymore? We can run on library code without main()? In context of C++ this sound better, not sure about other languages
```
... detect bugs in Java and C/C++/Objective-C code before it ships
```

jvillard: Sounds good!

PR #617, opened 2017-03-10
display additional options in help text

3flex:
These are mentioned on fbinfer.com so IMHO should be displayed when running infer -h


3flex:
Build failure was due to a timeout on one of the jobs.

PR #627, opened 2017-03-21
Update powered_by.yml

jrrexliang01:
add getsource and areimst

### DIFF
--- a/_data/powered_by.yml
+++ b/_data/powered_by.yml
@@ -10,3 +10,7 @@
     url: https://www.moneylover.me/
   - name: OLA
     url: https://www.olacabs.com/
+  - name: GetSource
+    url: https://www.getsource.cn/
+  - name: Areimst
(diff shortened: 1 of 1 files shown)


_data/powered_by.yml
```diff
@@ -10,3 +10,7 @@
     url: https://www.moneylover.me/
   - name: OLA
     url: https://www.olacabs.com/
+  - name: GetSource
+    url: https://www.getsource.cn/
```
jvillard: This doesn't load for me. Could you double-check the URL?

_data/powered_by.yml
```diff
@@ -10,3 +10,7 @@
...
+  - name: Areimst
+    url: https://www.areimst.com/
```
jvillard: This doesn't load either. Did you mean `http://areimst.com/`?

Issue #644, opened 2017-04-20
Cannot build Infer by `build-infer.sh` because of incompatible version with some packages in Opam


tddsg:
Hi there,

- I'm trying to build the newest version of Infer (https://github.com/facebook/infer/commit/bcbb0320521845e7bea0fda3812406ee0ec83bca), but encounter the following errors:

```
The following dependencies couldn't be met:
  - ppx_bin_prot -> ppx_type_conv < 113.34.00+4.03
Your request can't be satisfied:
  - ppx_type_conv<113.34.00+4.03 is not available because your system doesn't comply with ocaml-version = "4.02.3".

```

- This seems not right to me, since the default version of OCaml used by Infer is 4.04.0, according to `./build-infer.sh -h`:

```
Usage: ./build-infer.sh [-y] [targets]

 targets:
   all      build everything (default)
   clang    build C and Objective-C analyzer
   java     build Java analyzer

 options:
   -h,--help             show this message
   --only-install-opam   install the opam dependencies of infer and exists
   --opam-switch         specify the opam switch where to install infer (default: infer-4.04.0)
   -y,--yes              automatically agree to everything

 examples:
    ./build-infer.sh               # build Java and C/Objective-C analyzers
    ./build-infer.sh java clang    # equivalent way of doing the above
    ./build-infer.sh java          # build only the Java analyzer

```

- Here is the detail of my OCaml setting:

```
which ocaml
/home/trungtq/.opam/infer-4.04.0/bin/ocaml

```

- And here is the full log thrown by `build-infer.sh`:
```

./build-infer.sh 
initializing opam... 
OPAM has already been initialized.
=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

1. To configure OPAM in the current shell session, you need to run:

      eval `opam config env`

2. To correctly configure OPAM for subsequent use, add the following
   line to your profile file (for instance ~/.profile):

      . /home/trungtq/.opam/opam-init/init.sh > /dev/null 2> /dev/null || true

3. To avoid issues related to non-system installations of `ocamlfind`
   add the following lines to ~/.ocamlinit (create it if necessary):

      let () =
        try Topdirs.dir_directory (Sys.getenv "OCAML_TOPLEVEL_PATH")
        with Not_found -> ()
      ;;

[...]

Issue #643, opened 2017-04-19
Can't run build-infer.sh


adilek:
I can't run build-infer.sh due to some error related to reason-parser.

I do get error message like this:
```
...
∗  installed atd.1.2.0
∗  installed ppx_optcomp.113.33.01+4.03
∗  installed atdgen.1.10.0
∗  installed camomile.0.8.5
∗  installed ppx_driver.113.33.02+4.03
[ERROR] The compilation of reason-parser failed at "./build.native build --native true --native-dynlink true".
∗  installed ppx_let.113.33.00+4.03
∗  installed ppx_pipebang.113.33.00+4.03
∗  installed ppx_here.113.33.00
∗  installed zed.1.4
∗  installed ppx_type_conv.113.33.02+4.0
...
===== ∗  3 =====

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[reason] Archive in cache
[reason-parser] Archive in cache
[infer-deps.0.11.0] /home/adil/inferlatest/infer/dependencies/infer-deps/ already up-to-date

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] The compilation of reason-parser failed at "./build.native build --native true --native-dynlink true".

#=== ERROR while installing reason-parser.1.13.2 ==============================#
# opam-version 1.2.2
# os           linux
# command      ./build.native build --native true --native-dynlink true
# path         /home/adil/.opam/infer-4.04.0/build/reason-parser.1.13.2
# compiler     4.04.0
# exit-code    1
# env-file     /home/adil/.opam/infer-4.04.0/build/reason-parser.1.13.2/reason-parser-15990-6d07ae.env
# stdout-file  /home/adil/.opam/infer-4.04.0/build/reason-parser.1.13.2/reason-parser-15990-6d07ae.out
# stderr-file  /home/adil/.opam/infer-4.04.0/build/reason-parser.1.13.2/reason-parser-15990-6d07ae.err
### stdout ###
# [...]
# ocamlfind ocamldep -package 'menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend result dynlink findlib.dynload ocaml-migrate-parsetree ppx_tools_versioned.metaquot_404' -modules src/reason_parser_message.ml > src/reason_parser_message.ml.depends
[...]

PR #646, opened 2017-04-20
fixup to address the issues #644 and #643

adilek:
https://github.com/facebook/infer/issues/643
https://github.com/facebook/infer/issues/644


adilek:
Perhaps travis check will fail because docker uses the files from release and the opam file needs to be changed.

Issue #669, opened 2017-06-16
Capturing Gradle project does not work for the path with spaces


nuald:
**Scope**: either stable or the current development version
```
> infer/bin/infer --version
Infer version v0.12.0-572a383f
```
**STR** (assumes macOS, but it should be reproducible for all platforms):
```
> mkdir "$TMPDIR/a b"
> cp -r examples/android_hello/ "$TMPDIR/a b"
> infer/bin/infer -- "$TMPDIR/a b"/gradlew -b "$TMPDIR/a b/build.gradle" build
...
ERROR:
*** Failed to execute compilation command: exited with code 2
...
*** Output:
javac: file not found: b/app/src/main/java/infer/inferandroidexample/Generated.java
```
**Note**
The PR is under way, it's rather trivial issue to fix.

jeremydubreil:
@nuald: Yes, it is possible that we never encounter this case before. Thanks for trying to send a PR.

jvillard:
Fixed by c2f7ee9.

PR #670, opened 2017-06-16
Fixed the gradle capture for the projects with spaces in their path.

nuald:
A fix for the reported issue #669.


infer/lib/python/inferlib/capture/gradle.py
```diff
@@ -62,19 +81,38 @@ def get_infer_commands(self, verbose_output):
                 javac_arguments = content.split(' ')
                 java_files = []
                 java_args = []
+                collected = []
                 for java_arg in javac_arguments:
-                    if java_arg.endswith('.java'):
-                        java_files.append(java_arg)
-                    else:
+                    if java_arg and java_arg[0] == '-':
```
jvillard: This won't work for "SomeFileWith -ASpaceAndADash.java" for instance, or for files starting with a `-`.

nuald:
Thanks for the review, I'll update the PR. Few notes:

- I guess the more reliable way would be the parsing from the tail of the `javac_arguments` list to the head, it would cover the use case "this -silly name.java" (but it will require more intelligent `extract_filepath` function that would use the greedy algorithm to distinguish between options and file names);
 - I don't see the unit-tests for the python code (only some integration tests), is it ok to embed doctests into the file, or do you want me to use some specific testing routine?
 - I think that the actual **infer** plugin for Gradle would be more appropriate solution comparing with okbuck. I can take on this (it's not too much work, at least for the initial version - just a matter of reading source sets and Java compile options), but not sure how to proceed with that, because I guess all **infer** projects should belong to this repo (also it doesn't fit the current way of working with **infer**).

jeremydubreil:
@nuald: We don't have unit tests in Python and are using integration tests essentially because we are trying to progressively migrate the Python code to OCaml by relying on those integration tests. Maybe, renaming the directory `android hello` instead of `android_hello` would be enough.

nuald:
The PR has been updated:

- Dashed filenames should be working now;
- Doctests have been added.

Please note that I've changed few imports to be lazy (so they called inside the functions, not during the module imports) - it's required to simplify the doctests.

infer/lib/python/inferlib/capture/gradle.py
```diff
@@ -31,9 +29,92 @@ def gen_instance(*args):
...
+    >>> extract_all(['-opt1', 'opt', 'val1', '/path/to/1.java'])
+    {'files': ['/path/to/1.java'], 'opts': ['-opt1', 'opt val1']}
+    >>> extract_all(['-opt1', '/path/to/a', 'b/c', 'd/1.java', '-opt2'])
+    {'files': ['/path/to/a b/c d/1.java'], 'opts': ['-opt1', '-opt2']}
+    >>> extract_all(['-opt1', 'optval1', '-path/to/1.java'])
+    {'files': ['-path/to/1.java'], 'opts': ['-opt1', 'optval1']}
+    >>> extract_all(['-opt1', 'optval1', '/path/to/', '-1.java'])
+    {'files': ['/path/to/ -1.java'], 'opts': ['-opt1', 'optval1']}
+    >>> extract_all(['undef1', 'undef2'])
+    {'files': [], 'opts': ['undef1 undef2']}
```
jvillard: Is it a good idea to merge these two together?

infer/lib/python/inferlib/capture/gradle.py
```diff
@@ -31,9 +29,92 @@ def gen_instance(*args):
...
+    """
+    def pop():
+        if javac_arguments:
```
jvillard: We try to be consistent in specifying unambiguously what is tested in a conditional, for instance `if javac_arguments is not None:` instead of `if javac_arguments:`. Could you please update your PR to follow this convention? There are a few more occurrences of this pattern in the lines you added.

infer/lib/python/inferlib/capture/gradle.py
```diff
@@ -31,9 +29,92 @@ def gen_instance(*args):
...
+        if java_arg.endswith('.java'):
+            # Probably got a file
+            remainder, path = extract_filepath(javac_arguments + [java_arg])
+            if path:
+                java_files.append(path)
+                javac_arguments = remainder
+            else:
+                # A use-case here: *.java dir as an option parameter
+                rev_opt_params.append(java_arg)
+        elif java_arg[0] == '-':
```
jvillard: This cannot happen because Gradle doesn't escape anything it gives us, but better use `java_arg.startswith('-')` in case `java_arg` has `0` length, to avoid crashing.

nuald: All empty items are filtered before the function call, but surely I will modify the code for the consistency.

infer/lib/python/inferlib/capture/gradle.py
nuald: Sure, I'll do the explicit checks.

infer/lib/python/inferlib/capture/gradle.py
nuald: Frankly, I'm not sure what to do here. By the javac docs, besides options and java files we may have classes and @argfiles, for the former it's better to split, for the latter - join. I think it's better to improve the logic, and assume that if we don't have any @argfiles, the leftovers are the names of the classes. It will be addressed in the up-to-dated PR.

jvillard: To be honest this can only be a best-effort attempt and it's hard to know which one will be best in practice so I'm fine with whichever. I wouldn't spend too much time on this. As you mentioned, time would be better spent writing a better integration from scratch using a plugin, or find a way to extract some sort of compilation database from Gradle and create a build-system-neutral interface in infer to read these that could be re-used for other build systems for which infer doesn't have a good integration.

jvillard:
Thanks a lot @nuald!

jberdine:
Unfortunately we have had to [revert](https://github.com/facebook/infer/commit/8c6615963f79f03a644ae9087eb160da89a09a1a) this for now since in practice, extract_filepath exhausts the call stack, leading to "RuntimeError: maximum recursion depth exceeded" exceptions. Perhaps a loop needs to be used instead.

Issue #676, opened 2017-06-27
--java-jar-compiler is undocumented


rgrig:
The option `--java-jar-compiler` is undocumented.

Is it intended to stay? Or should I not rely on it?

jvillard:
It should be reliable for the time being at least. It's hidden in `--help-full` at the moment but it shouldn't be. It should be easy to move it to `infer capture --help` too (one would need to add `~in_help:CLOpt.[Capture, manual_java]` to [this function call](https://github.com/facebook/infer/blob/a8ad84b9d3c14734387dc8715818eaae0b42c8f2/infer/src/base/Config.ml#L1167)):

```
       --java-jar-compiler path
           Specifify the Java compiler jar used to generate the bytecode
```

jvillard:
Closed by 8c98bee

PR #677, opened 2017-06-28
Make --java-jar-compiler visible in `infer capture --help`.

rgrig:
made change suggested by @jvillard in https://github.com/facebook/infer/issues/676


jvillard:
Thanks @rgrig :)

PR #682, opened 2017-07-04
add cursorClosedCheckNullCheckClosed test

azhdar:


infer/tests/codetoanalyze/java/infer/CursorLeaks.java
```diff
@@ -53,6 +53,26 @@ public Object cursorClosedCheckNull(SQLiteDatabase sqLiteDatabase) {
     }
     return value;
   }
+  
+  public Object cursorClosedCheckNullCheckClosed(SQLiteDatabase sqLiteDatabase) {
```
jvillard: Infer reports a resource leak on this method, because it doesn't know about `isClosed()`. We can add this test to the codebase by merging your PR, then fix the modeling issue separately. Let me know if you'd like to give it a go! It probably involves modeling `isClosed()` in our [model of cursors](https://github.com/facebook/infer/blob/2ee8321456cf3a26adb732d71f7bac534baf1b45/infer/models/java/src/android/database/Cursor.java).

If you don't mind I'll change the name of this method to `cursorClosedCheckNullCheckClosed_FP` and record the new test output since otherwise this breaks `make test` because of the newly reported error (see the [testing section of our CONTRIBUTING guide](https://github.com/facebook/infer/blob/master/CONTRIBUTING.md#testing-your-changes)).

PR #688, opened 2017-07-10
Fix javac wrapper

ratatosk:
1. Allow it to forward -version calls to javac (so that if the build system wants 
    to check the compiler version it doesn't end up with Infer's version)
2. Preserve empty arguments:
    javac -bootclasspath "" -classpath a.jar ...
    currently gives:
    javac: invalid flag: a.jar

### DIFF
--- a/infer/lib/wrappers/javac
+++ b/infer/lib/wrappers/javac
@@ -11,11 +11,13 @@ elif [ -z "$INFER_OLD_PATH" ]; then
 fi
 
 HOST_COMPILER=(`PATH=$INFER_OLD_PATH which javac`)
-COMPILER_ARGS="$@"
-HOST_COMPILER_COMMAND=("$HOST_COMPILER" $COMPILER_ARGS)
-FRONTEND_COMMAND=("infer" "-a" "capture" "-o" "$INFER_RESULTS_DIR" "--" "javac" $COMPILER_ARGS)
+COMPILER_ARGS=("$@")
+HOST_COMPILER_COMMAND=("$HOST_COMPILER" "${COMPILER_ARGS[@]}")
+FRONTEND_COMMAND=("infer" "-a" "capture" "-o" "$INFER_RESULTS_DIR" "--" "javac" "${COMPILER_ARGS[@]}")
 
-if [ -n "$INFER_COMPILER_WRAPPER_IN_RECURSION" ]; then
+if [[ "$*" == *-version* ]]; then
+    "${HOST_COMPILER_COMMAND[@]}"
+elif [ -n "$INFER_COMPILER_WRAPPER_IN_RECURSION" ]; then
     if [ -z "$INFER_LISTENER" ]; then
         "${HOST_COMPILER_COMMAND[@]}"
     fi


jeremydubreil:
Thanks for the PR. Did you see an example where the `-version` case was breaking something?

ratatosk:
Jeremy, yeah, Android (OS itself) build uses -version to check if the correct javac is used.
https://android.googlesource.com/platform/build/soong/+/master/ui/build/java.go#54

jvillard:
Thanks again @ratatosk. What is the toplevel command you used to run infer that ended up using this wrapper? I noticed a few more improvements to make and I'd like to make sure I won't introduce regressions.

ratatosk:
@jvillard, the top-level command was this (in a docker container with some additional packages), it is supposed to compile all java code of Android (the details can be found here: https://source.android.com/source/building).
`ANDROID_JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64 INFER_OLD_PATH=$PATH INFER_RESULTS_DIR=/infer-out infer capture -o /infer-out -- make ANDROID_COMPILE_WITH_JACK:=false java
`
(It is time consuming and requires some setup)
The variables are needed because apparently they aren't supplied if javac is called indirectly (not by make itself), IIUC.

Actually it ended up only capturing part or the java sources that were compiled for some reason, haven't investigated thoroughly. And when I attempted compilation with -j it failed with not a very clean error. Perhaps that happens because it's not just make but also uses soong now? I'll try to investigate later, but not sure exactly when since the build is slow and time consuming.

I got better capturing results with a more ad hoc solution, I put this wrapper (similar to yours one) in place of the real javac binary:
```
#!/usr/bin/python
import os, sys

javac_path = '/usr/lib/jvm/java-8-openjdk-amd64/bin/javac.real'
infer_path = '/infer/infer/bin/infer'

if "-version" in sys.argv:
    print('javac 1.8.0_131')

elif "IN_INFER" in os.environ:
    os.execv(javac_path, sys.argv)

else:
    env = os.environ.copy()
    env["IN_INFER"] = 'true'
    args = [infer_path, 'capture', '--reactive', '--'] + sys.argv
    os.execve(infer_path, args, env)
```
I wonder whether this approach should work in general for custom build systems and whether it allows Infer to get all possible information about the source?
And how to analyze files captured this way? When I run `infer analyze --reactive --changed-files-index index.txt` for some list of files, it returned nothing (although it showed a lot of Fs and dots while running). But when I ran `infer --reactive --changed-files-index index.txt -- analyze` for the same files it showed some issues.
Is the latter way of invocation (with 'analyze' as compile command) documented anywhere? I found it in some github discussion.

Thanks!

jvillard:
Cool, many thanks for that write-up! I have an upcoming diff that should address the issues you've encountered.

> The variables are needed because apparently they aren't supplied if javac is called indirectly (not by make itself), IIUC.

This was indeed broken.

> it ended up only capturing part or the java sources that were compiled for some reason

I noticed that the bash wrapper didn't set `--continue` so would wipe out infer-out/ on every `javac` invocation. What you end up with is the results of only the last `javac` call.

> And when I attempted compilation with -j it failed with not a very clean error. 

No idea about that one, sorry.

> When I run infer analyze --reactive --changed-files-index index.txt for some list of files, it returned nothing

Which version of infer did you use? I think this was fixed by 84b9901b2270b728bbd191c925ba6b4d3c8544f6. `infer -- analyze` is the old way of calling `infer analyze`.

Your wrapper looks like it should work! I'm fixing the bash wrapper, it should work just as well after that.

ratatosk:
Great, I'll try again after the fixes and will come back if discover any discrepancy between those ways of invocation. Thanks!

PR #694, opened 2017-07-18
Flush output after printing bugs in txt format.

rgrig:
If you run `infer report --issues-txt ilovecats.txt ...` then bugs may mysteriously miss from `ilovecats.txt`, unless you flush. (See rules of thumb for `Format` module.)


jvillard:
Thanks! Sorry it took so long to merge.

PR #697, opened 2017-07-21
Updated fix for the gradle capture issue regarding projects with space-containing paths

nuald:
Fix for the #669. Recursion problem has been reproduced (see the new doctest in the latest commit) and eliminated. I guess the new version will work faster too (no more unnecessary slicing).


jvillard:
Apologies for the tardy update. This change slows down infer capture by as much as 40% on some large gradle project. Do you think that there are some low hanging fruits that could improve the performance of this change? If not, we could hide this behaviour behind a config flag (I can do that).

jvillard:
A thought: we could "parse" the javac command line given by gradle assuming there's no whitespace in paths, then check that all the .java files that we've found exist. If yes, job done, and if not, then run the slower parsing. Let me see if that helps, I'll report back.

infer/lib/python/inferlib/capture/gradle.py
```diff
@@ -31,9 +30,136 @@ def gen_instance(*args):
 create_argparser = util.base_argparser(MODULE_DESCRIPTION, MODULE_NAME)
 
 
+def extract_filepath(parts):
+    pos = 0
```
jvillard: Simply reversing the order of the iteration here solves the perf regression issue. No need to update your PR (unless you want to!), I can do the change on my end.

```Python
def extract_filepath(parts):
    size = len(parts)
    pos = size - 1
    while pos >= 0:
        path = ' '.join(itertools.islice(parts, pos, None))
        if os.path.isfile(path):
            return parts[:pos], path
        pos -= 1
    return parts, None
```

nuald: Please do. I guess the original order has been used to avoid use-cases like colliding "path to file.java" and "path/ to file.java" (substring of the first path is a directory for the second one). However, tests do not fail and if it works for you, I doubt we should care about minor use-case (if it's even possible, considering Javac requirements for the file names).

jvillard: Great, thanks for the context. I agree that we shouldn't worry about use-cases that are purely hypothetical for now.

jvillard:
Thanks again @nuald!

PR #716, opened 2017-08-11
Add tty debug

gregory-nisbet:
really simple `tty_debug` function that takes a string and writes it to the tty. Infer redirects stdout and stderr internally under some circumstances. This function is only intended to be used for `printf`-style debugging (hopefully the comment makes that clear).

### DIFF
--- a/infer/src/base/Logging.ml
+++ b/infer/src/base/Logging.ml
@@ -236,6 +236,20 @@ let debug kind level fmt =
   let to_file = compare_debug_level level base_level <= 0 in
   log ~to_console:false ~to_file debug_file_fmts fmt
 
+(* file descriptor for tty, lazily allocated *)
+let dev_tty = ref None;;
+
+let tty_debug str =
+  (match (!dev_tty) with
+  | None -> dev_tty := Some (Out_channel.create "/dev/tty")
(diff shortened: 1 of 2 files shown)


jvillard:
Sorry for the late reply. It would be good to have something like what you propose. `Logging.progress` always prints on stderr too, but it's not very discoverable due to the name.

I think it would be better for the new function to be implemented similarly to `Logging.progress` (or even be a simple alias to it) to be able to take a format string as argument. I'd also prefer a name like `Logging.dev_debug` to make it clearer that it can be used for debugging in development. If you add a tag `[@@deprecated "don't commit this, use Logging.debug instead"]` in the .mli it will make sure that uses do not get committed in the repo.

Could you also update the PR to remove changes to the facebook-clang-plugins submodule?

mbouaziz:
What's the status of this PR?

jvillard:
The idea of this PR is good: it would be useful to have a way to send things to the console without fearing that the output will get swallowed somehow. However, I would like something more involved than what this PR was going for originally, otherwise nothing will guarantee that leftover debugging instructions do not make it into the code base. I think we can close it and I'll send a diff implementing what I suggested.

PR #742, opened 2017-09-05
use sudo to run the linux script

mauryat:


jvillard:
Sorry for the late reply. Can you say more about why the script should be run as root?

jeremydubreil:
@mauryat, any update on why you need to the script to run as root?

mauryat:
@jeremydubreil , sorry I've not had a chance to look at it again as this was a long time ago. But, from what I remember, Ubuntu Linux required sudo permission to be able to run the script.

jvillard:
Let's reopen this once we find out why/when using `sudo` is needed.

PR #792, opened 2017-11-03
Adjust Dockerfile to build 0.13

aaronjwood:
I was unable to build 0.13 with the current Dockerfile. Changing the Infer version to 0.13 resulted in errors about the ocaml version not being >= 4.05 and < 4.06. When changing the ocaml version I found that the sqlite extension wouldn't compile via opam.

I tried to clean this up a little bit as I went but please let me know if I have reduced it too much. I made sure that everything still worked by running following the Docker steps outlined here http://fbinfer.com/docs/getting-started.html

Here's some info about the bits in the finalized container:

```
infer --version
Infer version v0.13.0
Copyright 2009 - present Facebook. All Rights Reserved.
```

```
opam --version
1.2.2
```

```
ocaml -version
The OCaml toplevel, version 4.02.3
```

```
java -version
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-2ubuntu1.17.04.3-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -29,8 +29,7 @@ RUN INFER_VERSION=v0.13.0; \
     ln -s ${PWD}/infer-linux64-$INFER_VERSION /infer
 
 # Compile Infer
-RUN OCAML_VERSION=4.05.0; \
-    opam init --compiler=$OCAML_VERSION -j $(nproc) --yes && \
+RUN OCAML_VERSION=4.05.0+flambda; \
     cd /infer && ./build-infer.sh --opam-switch $OCAML_VERSION && rm -rf /root/.opam
 
 # Install Infer
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,37 +1,22 @@
-# Base image
-FROM buildpack-deps:xenial-curl
+FROM ubuntu:17.04
 
 MAINTAINER Infer
 
-# Debian config
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
+    apt-get install -y \
             aspcud \
-            autoconf \
-            automake \
-            gcc \
-            g++ \
-            git \
-            groff-base \
-            libc6-dev \
-            libgmp-dev \
-            libmpc-dev \
-            libmpfr-dev \
-            m4 \
-            make \
-            ncurses-dev \
-            ocaml \
+            curl \
             opam \
-            openjdk-8-jdk-headless \
             pkg-config \
-            python2.7 \
-            rsync \
-            unzip \
-            zlib1g-dev && \
+            autoconf \
+            libsqlite3-dev \
+            zlib1g-dev \
+            cmake \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -40,25 +29,14 @@ RUN INFER_VERSION=v0.12.1; \
...
-    rm -rf /root/.opam
+RUN OCAML_VERSION=4.05.0; \
+    opam init --compiler=$OCAML_VERSION -j $(nproc) --yes && \
+    cd /infer && ./build-infer.sh --opam-switch $OCAML_VERSION && rm -rf /root/.opam
 
 # Install Infer
 ENV INFER_HOME /infer/infer
 ENV PATH ${INFER_HOME}/bin:${PATH}
 
-# Install dependencies for Android sample
```
jvillard: This used to be installed separately, but I'm not sure if this matters much. @3flex?

docker/Dockerfile
```diff
@@ -40,25 +29,14 @@ RUN INFER_VERSION=v0.12.1; \
     ln -s ${PWD}/infer-linux64-$INFER_VERSION /infer
 
 # Compile Infer
-RUN OCAML_VERSION=4.04.2; \
-    opam init --compiler=$OCAML_VERSION -j $(getconf _NPROCESSORS_ONLN || echo 1) --yes && \
-    cd /infer && \
-    ./build-infer.sh --opam-switch $OCAML_VERSION && \
-    rm -rf /root/.opam
+RUN OCAML_VERSION=4.05.0; \
```
jvillard: Building infer with flambda makes its backend ~25% faster but worsen build times (f8d7c810452ce3a4d2e7027e38f5d00426a2a917). I think it's worth turning on by default, like we do in the in-tree build:
```
RUN OCAML_VERSION=4.05.0+flambda; \
```

docker/Dockerfile
```diff
@@ -40,25 +29,14 @@ RUN INFER_VERSION=v0.12.1; \
...
+    opam init --compiler=$OCAML_VERSION -j $(nproc) --yes && \
```
jvillard: Not needed anymore I think.

docker/Dockerfile
aaronjwood: Sure, will fix this.

docker/Dockerfile
aaronjwood: My thought behind this was that we could do it all in one shot when the initial `apt-get update && apt-get install` is done. Also saves a layer :)

igrep:
I'm also interested in this pull request.
It's tough to build infer without Docker!

jvillard:
Apologies for taking a bit long to merge this. Thanks again for the fix!

aaronjwood:
No problem!

Issue #797, opened 2017-11-07
Recent versions of infer fail to run on WSL (Windows Subsystem for Linux)


vsilyaev:
Recent version of infer (with the SQLite backed ResultsDatabase) fail to WSL, which manifests itself in random crashes like:
```
[*ERROR**][32701] Makefile:43: recipe for target '/mnt/c/my/work/infer/infer/lib/specs/clang_models' failed
[*ERROR**][32701] make[1]: Leaving directory '/mnt/c/my/work/infer/infer/models'
[*ERROR**][32701] *** stderr:
[*ERROR**][32701] Capturing in make/cc mode...
[*ERROR**][32701] Capturing in make/cc mode...
[*ERROR**][32701]
[*ERROR**][32701] Error backtrace:
[*ERROR**][32701] Raised at file "base/SqliteUtils.ml" (inlined), line 15, characters 46-63
[*ERROR**][32701] Called from file "format.ml", line 1275, characters 4-36
[*ERROR**][32701] Called from file "base/SqliteUtils.ml" (inlined), line 29, characters 20-52
[*ERROR**][32701] Called from file "base/ResultsDatabase.ml" (inlined), line 110, characters 2-75
[*ERROR**][32701] Called from file "base/ResultsDir.ml" (inlined), line 49, characters 2-44
[*ERROR**][32701] Called from file "backend/infer.ml" (inlined), line 43, characters 6-38
[*ERROR**][32701] Called from file "backend/infer.ml", line 83, characters 2-10
[*ERROR**][32701] Uncaught error: ("InferModules__SqliteUtils.Error(\"exec: synchronous=OFF: IOERR\")")
[*ERROR**][32701]
[*ERROR**][32701] Run the command again with `--keep-going` to try and ignore this error.
[*ERROR**][32701] make[4]: *** [glib.o] Error 2

infer/bin/infer --version
Infer version v0.13.0-c542b65
Copyright 2009 - present Facebook. All Rights Reserved.
```
Most likely it's caused by partial implementation of the POSIX interfaces (i.e. bug) in the WSL itself. See:
https://github.com/Microsoft/WSL/issues/1927
https://github.com/Microsoft/WSL/issues/2395

P.S. Infer v.012 works just fine on WSL

PR #798, opened 2017-11-08
Use SQLite in a way that is compatible with current state of WSL

vsilyaev:
To resole #797  this adds runtime option to select VFS for SQLite, 

When infer runs on WSL this defaults to "unix-excl" (https://sqlite.org/vfs.html) and if VFS is specified, then   WAL  is not enabled (since WAL is non compatible with custom VFS  - https://www.sqlite.org/wal.html).

### DIFF
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -1874,6 +1874,19 @@ and specs_library =
   specs_library
 
 
+and sqlite_vfs =
+  let default =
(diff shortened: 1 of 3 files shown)


jvillard:
This looks great, many thanks for fixing this.

jvillard:
Would you mind signing our CLA? I would very much like to merge this. https://code.facebook.com/cla

PR #801, opened 2017-11-09
[C/C++] Encode filename into the Procname of global initializers

### LINKED ISSUE #796: [C/C++] Failure to analyze all static initializers
It seems that if project has several static variables  with the same name, infer will analyze only one of them.
There is an example:
```C
>cat >test1.cpp <<EOF
int div(int a) {
    return 1/a;
}

static int init = div(0);
EOF
>cat >test2.cpp <<EOF
int div2(int a) {
    return 1/a;
}

static int init = div2(0);
EOF
```
```
>infer --no-filtering -- c++ -c test1.cpp test2.cpp
Capturing in make/cc mode...
Found 2 source files to analyze in /tmp/infer-out
Starting analysis...

legend:
  "F" analyzing a file
  "." analyzing a procedure

FF...

Found 1 issue

test2.cpp:5: error: DIVIDE_BY_ZERO
  Expression `0` could be zero at line 5, column 19.
  3.   }
  4.
  5. > static int init = div2(0);


Summary of the reports

  DIVIDE_BY_ZERO: 1
>infer --version
Infer version v0.13.0-ff475e4
Copyright 2009 - present Facebook. All Rights Reserved.
```
It appears that  all global initializers  for 'init' were mapped to entry with the same name, so only one(last) survives.
```
>ls -1 infer-out/specs/
div2#17260766888000402206.597b37a95a09a586d7a131a160dbcaa9.specs
div#939704114457859019.e5abef53fc8ff782b87b6a9041fd5e9a.specs
__infer_globals_initializer_init.e23415c599c5045a2828825332adee29.specs
````

vsilyaev:
This resolves #796 . Effectively it adds file specific suffix to name of all global initializers (so initializersof two global variable of the same name will have unique Typ.Procname). which is the same rule as currently used by constructing Procname for the static functions.  However this change applies to initializers of all global variables and not just static (arguably it's a right thing. since GCC used to allow multiple global variables with the same name).

Consequences of this change that it becomes impossible to know name of generated initialization function of global ('extern') variables. However get_initializer_pname function is only referenced by the frontend (when creating initializer for the defined global  variables) and by the SIOF checker.

### DIFF
--- a/infer/src/IR/Pvar.ml
+++ b/infer/src/IR/Pvar.ml
@@ -277,8 +277,6 @@ let get_initializer_pname {pv_name; pv_kind} =
   match pv_kind with
   | Global_var (translation, _, _, _, is_static_global) ->
       let name = Config.clang_initializer_prefix ^ Mangled.to_string_full pv_name in
-      if is_static_global then print_endline ("Static : " ^ name) ;
-      (* let is_static_global = false in *)
       if is_static_global then
         match translation with
         | TUFile file ->
--- a/infer/src/IR/Pvar.ml
+++ b/infer/src/IR/Pvar.ml
@@ -24,9 +24,9 @@ type pvar_kind =
       (** synthetic variable to represent return value *)
   | Abduced_ref_param of Typ.Procname.t * int * Location.t
       (** synthetic variable to represent param passed by reference *)
-  | Global_var of (translation_unit * bool * bool * bool)
+  | Global_var of (translation_unit * bool * bool * bool * bool)
       (** global variable: translation unit + is it compile constant? + is it POD? + is it a static
-      local? *)
+      local?  + is it a static global *)
   | Seed_var  (** variable used to store the initial value of formal parameters *)
   [@@deriving compare]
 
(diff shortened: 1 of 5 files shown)


jvillard:
Does this break the SIOF checker then?

I'm in the process of moving these objects (CFGs, specs, ...) into the SQLite database, this can probably help if your solution is at the expense of SIOF. For instance, we'd be able to query for the initializer of a global without having to have the same name for all of them like we do now.

vsilyaev:
> Does this break the SIOF checker then?
Yes, it was broken. Now updated fix will only affect names of static (file scope) initializers, and will not affect initializers with global scope. So SIOF passes with latest changes (since it seems to not use file scoped globals.

However this change is more involved and requires some changes in facebook/facebook-clang-plugins#5  so, it'd be possible to differentiate variables of the global and file(global) scope.

jvillard:
Thanks, sounds like a good fix! Would you mind signing the CLA that the bot mentioned so we can merge your changes?

dulmarod:
Hi @vsilyaev, I imported your pull request but our CI fails because the commit hash from facebook-clang-plugins doesn't exist. If you have a commit there, you need to commit it first, and then the Infer CI will work.

jvillard:
@dulmarod  https://github.com/facebook/facebook-clang-plugins/pull/5 has to be merged beforehand.

dulmarod:
Hi @vsilyaev, I ran CI again after merging you pull request from facebook-clang-plugin and now the Infer tests are failing. You should run `make direct_tests`. It's also worth rebasing to the latest master.

vsilyaev:
Hi @dulmarod,
direct_cpp_frontend tests were rerecorded and branch was rebased while also collapsing fixups.

dulmarod:
Hi, The code looks good now, but the tests are failing again, could you re-record them again? Thanks.

dulmarod:
Never mind, I fixed the tests and landed.

PR #816, opened 2017-11-20
Clean temporary files created by the clang frontend

vsilyaev:
This commit augments codes to clean up temporary files generated by the clang frontend.

Currently clang frontend leaves large number of temporary files int the tmp directory, and it could be seen for example by running these command:
```
git clean -xdf infer/models/
mkdir /tmp/infer
env TMPDIR=/tmp/infer make infer_models
ls -l /tmp/infer
```
P.S. Analyzing real project could easily cause each infer capture to leave hundreds of  files in /tmp
P.P.S.  There are 11 total references to Filename.temp_file however, other 9 seems don't leak temporary files in such large scale (at least not when using the clang frontend).

### DIFF
--- a/infer/src/base/CommandLineOption.ml
+++ b/infer/src/base/CommandLineOption.ml
@@ -965,6 +965,8 @@ let parse_args ~usage initial_action ?initial_command args =
   parse_loop () ; curr_usage
 
 
+let keep_args_file = ref false
+
 let parse ?config_file ~usage action initial_command =
   let env_args = decode_env_to_argv (Option.value (Sys.getenv args_env_var) ~default:"") in
   let inferconfig_args =
@@ -1008,6 +1010,7 @@ let parse ?config_file ~usage action initial_command =
          to prevent this from happening *)
       let file = Filename.temp_file "args_" "" in
       Out_channel.with_file file ~f:(fun oc -> Out_channel.output_lines oc argv_to_export) ;
+      if not !keep_args_file then Utils.unlink_file_on_exit file ;
       "@" ^ file
     else ""
   in
--- a/infer/src/base/ProcessPoolState.ml
+++ b/infer/src/base/ProcessPoolState.ml
@@ -1,0 +1,2 @@
+(** Keep track of whether the current execution is in a child process *)
+let in_child = ref false
--- a/infer/src/integration/ClangQuotes.ml
+++ b/infer/src/integration/ClangQuotes.ml
@@ -37,4 +37,5 @@ let mk_arg_file prefix style args =
   in
   Utils.with_file_out file ~f:write_args |> ignore ;
   L.(debug Capture Medium) "Clang options stored in file %s@\n" file ;
+  if not Config.debug_mode then Utils.unlink_file_on_exit file ;
   file
--- a/infer/src/base/CommandLineOption.mli
+++ b/infer/src/base/CommandLineOption.mli
@@ -249,3 +249,5 @@ val show_manual :
 (** Display the manual of [command] to the user, or [command_doc] if [command] is None. [format] is
     used as for [Cmdliner.Manpage.print]. If [internal_section] is specified, add a section titled
     [internal_section] about internal (hidden) options. *)
+
+val keep_args_file : bool ref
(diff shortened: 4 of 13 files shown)


infer/src/base/CommandLineOption.ml
```diff
@@ -990,6 +990,7 @@ let parse ?config_file ~usage action initial_command =
          to prevent this from happening *)
       let file = Filename.temp_file "args_" "" in
       Out_channel.with_file file ~f:(fun oc -> Out_channel.output_lines oc argv_to_export) ;
+      Pervasives.at_exit (fun () -> try Unix.unlink file with _ -> ()) ;
```
jvillard: Please use `Epilogues.register` so that this also works when infer is interrupted by a signal (unless this creates a dependency cycle?).

infer/src/integration/ClangQuotes.ml
```diff
@@ -37,5 +37,6 @@ let mk_arg_file prefix style args =
   in
   Utils.with_file_out file ~f:write_args |> ignore ;
   L.(debug Capture Medium) "Clang options stored in file %s@\n" file ;
+  Pervasives.at_exit (fun () -> try Unix.unlink file with _ -> ()) ;
```
jvillard: Use `Epilogues` here too.

vsilyaev:
@jvillard , thank you for reviewing.
> Should  removing them be gated by Config.debug, though? It's often useful to look at them when  something goes wrong.

Yes, contents of temporary file could be precious for debugging purposes. 0636027  will keep clang_xxx temp file while in the debug mode. However args_xxx file that are created by the  CommandLineOption still will be deleted unconditionally since there is no easy access to Config.debug_mode .

mbouaziz:
Thanks @vsilyaev for the PR!
- The args files are still useful for debugging purposes (another possibility could be to dump their content in the log)
- Could you rename `clean_temp_file` so that we understand it's not happening now?
- Why the reference, aren't we sure the epilogue is run only once? (Worst case: `unlink` fails)
@jvillard any idea why Travis CI is not happy?

vsilyaev:
@mbouaziz 
> * The args files are still useful for debugging purposes (another possibility could be to dump their content in the log)

b925a37 will keep args files in the debug mode. I wish there was better way of knowing whether infer runs in the debug mode, but it should work.

> * Could you rename clean_temp_file so that we understand it's not happening now?

2a53319 , is it better? 

> * Why the reference, aren't we sure the epilogue is run only once? (Worst case: unlink fails)

Well it was a knee jerk reaction on possibility of calling this function in response to signal, since signal could be caught again while doing slow unlink (for example unlinking NFS file). However I don't thing it's really worth it, so it's got removed by  46d0dc6 .

infer/src/base/CommandLineOption.ml
```diff
@@ -440,6 +443,13 @@ let mk_bool ?(deprecated_no= []) ?(default= false) ?(f= fun b -> b) ?(deprecated
   in
   let default_to_string _ = "" in
   let mk_spec set = Unit (fun () -> set "") in
+  let f =
+    (* also set local config_debug *)
+    if String.equal long "debug" then fun b ->
```
jvillard: I think this logic would better live in `Config`, just because then you don't need to compare to the `"debug"` string: you can rename `config_debug` above to `debug`, export it in CommandLineOption.mli, and set it from `Config` in `debug`'s [callback function](https://github.com/facebook/infer/blob/b71d56715729dcf982c53471fc45373cf59d016d/infer/src/base/Config.ml#L1058-L1060).

The only code duplication then is the default value of `false`.

infer/src/base/ProcessPoolState.ml
```diff
@@ -0,0 +1,2 @@
+(** Keep track of whether the current execution is in a child process *)
```
jvillard: Nice!

vsilyaev:
I've re-based branch  to tip of master, and resolved Config and CommandLineOption circular dependencies by adding mutable reference to the CommandLineOption ,

jvillard:
Thanks @vsilyaev!

PR #833, opened 2017-12-22
Fixed namespace for Java globals.

### LINKED ISSUE #831: Globals treated differently depending on whether they are in different files.
Put the following code in a file `Repro.java`

```Java
class P {
  static public Object x;
  public static void f() {
    if (x != null) x.hashCode();
  }
}

class C {
  void c() {  P.f(); }
}

public class Repro {
  static void g(C y) {
    if (y == P.x) {} else {}
    y.c();
  }
}
```
and run `infer -g -- javac Repro.java` with version d83a944. The specs for `g()` look fine:
ok

Now move class `P` to file `P.java`, clean the cache, and run again the same command. The specs for `g()` look very strange:
nok

rgrig:
In Java, static variables are distinguished by package/class:
the file where they are defined doesn't matter.

Fixes #831.

### DIFF
--- a/infer/src/IR/Pvar.ml
+++ b/infer/src/IR/Pvar.ml
@@ -16,7 +16,6 @@ module F = Format
 
 type translation_unit =
   | TUAnonymous
-  | TUExtern
   | TUFile of SourceFile.t
   [@@deriving compare]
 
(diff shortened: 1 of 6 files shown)


sblackshear:
Would it be possible to combine `TUExtern` and `TUAnonymous`? It seems like these are playing the same role except that one is used in Java and the other in Clang. `TUAnonymous` seems more appropriate for both than `extern` does.

rgrig:
Done.

(I rebased because of some unrelated OPAM strangeness.)

sblackshear:
Merged, thanks!

PR #851, opened 2018-01-25
eradicate-warnings docs: Fix source samples

passy:
There were some superfluous backticks that broke the rendering.

http://fbinfer.com/docs/eradicate-warnings.html

screenshot from 2018-01-25 12-12-53


jvillard:
Thanks for fixing this.

PR #861, opened 2018-02-07
Also detail the compilation database usage

sylvestre:


jvillard:
Thank you @sylvestre

sylvestre:
Happy to help!

PR #863, opened 2018-02-08
infer-ant: adapt the situation when too much files need to be compiled.

yalechen-cyw3:


infer/lib/python/inferlib/jwlib.py
```diff
@@ -24,14 +24,46 @@ def __init__(self, javac_args):
     def start(self):
         infer = os.path.join(config.BIN_DIRECTORY, 'infer')
         # pass --continue to prevent removing the results-dir
+        new_javac_args = []
+        list_count = 0
+        file_count = 0
+        source_list = "infer_sourcefiles"
+        if os.path.exists(source_list+str(list_count)):
```
jvillard: Please use `tempfile.NamedTemporaryFile()` instead, similar to [here](https://github.com/facebook/infer/blob/690bae30d0b6163f23075bf22fa5b4faf248104f/infer/lib/python/inferlib/capture/gradle.py#L196-L202) for instance.

infer/lib/python/inferlib/jwlib.py
```diff
@@ -24,14 +24,46 @@ def __init__(self, javac_args):
...
+            os.remove(source_list+str(list_count))
+        f = open(source_list+str(list_count), "a")
+        for arg in self.javac_args:
+            if arg.endswith(".java"):
+                file_count += 1
+                f.write(arg + "\n")
+                if file_count >= 1000:
```
jvillard: If I understand correctly, this is putting the arguments that are "*.java" into files, with no more than 1000 of these per argument file.

I wonder if this is needed: could you just put all the arguments inside the same temporary file instead? That would create less churn on the filesystem and be much simpler.

infer/lib/python/inferlib/jwlib.py
```diff
@@ -24,14 +24,46 @@ def __init__(self, javac_args):
...
             infer,
             'capture',
             '--continue',
             '--', 'javac'
         ] + self.javac_args
         try:
-            return subprocess.check_call(cmd)
+            # return subprocess.check_call(cmd)
+            exitcode = 0
+            for i in range(list_count+1):
```
jvillard: Why not run a single compilation command on all the .java files at once?

infer/lib/python/inferlib/jwlib.py
yalechen-cyw3: ok

infer/lib/python/inferlib/jwlib.py
yalechen-cyw3: I will try it.

jvillard:
More updates were needed here. We are working on rewriting the `ant` integration in OCaml too so the PR won't apply anymore (but the new integration will likely have whatever problem this was addressing).

PR #872, opened 2018-02-26
Detecting strong ivars

Megra:
This pull-request brings the capability to check the lifetime of ivars, to warn about types that should not be retained strongly. To do so I had to improve has_type() to support AttributeType and added a new function to check the lifetime attached to a pointer.

I added a test to verify both work as expected.

Sanity checks I ran:

  make -j test_build && make test

I ran clang-format on the modified files but it lead to massive changes with dubious indentation so I have not included these changes here. Let me know if you think i did it wrong and should do so. I ran clang-format from the root of the repository, where .clang-format is stored in.

### DIFF
--- a/infer/src/clang/cPredicates.ml
+++ b/infer/src/clang/cPredicates.ml
@@ -716,6 +716,42 @@ let has_type_const_ptr_to_objc_class node =
   | None ->
       false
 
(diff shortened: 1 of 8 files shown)


jvillard:
Closed by 8cfc7f2cada7c2e3c37f629aa1c34cd0e8f8de28

PR #1042, opened 2019-01-04
WIP Depend on Core_kernel instead of Core as much as possible

yakobowski:
The goal of this pull request is to have Infer depend on `Core_kernel` instead of `Core`, in order to facilitate its porting to windows. My approach was to introduce a compatibility layer that adds, on top of `Core_kernel`, the `Core` APIs that Infer uses. The goal is not to be exhaustive; on the contrary, we want to port as little of the code as possible, given the size of `Core minus Core_kernel`.

This pull request is still very much a work-in-progress, but I wanted to discuss beforehand if something of that kind could be merged. If yes, I would probably introduce a configure-time switch that would enable either the compatibility layer (which I would removed from `IStd`), or the entire Core codebase. We at AdaCore would be responsible for maintaining the compatibility layer of course. 

I would like feedback on a few points:
- the general approach itself :smiling_imp: 
- if possible suggestions of what should be done for the remaining "hard" dependencies, namely `base/{Process,ProcessPool}`
- inputs on two of the TODOs I left within the compatibility layer
  - using `sigkill` instead of `sigterm`
  - is the use of `Unix.flock` really important for day-to-day use? It seems to be used in `Utils.write_file_with_locking`, itself used by `PerfStats`

My next step will be to attempt to compile this version under Windows.

Thanks!


infer/src/bufferoverrun/bufferOverrunDomainRelation.ml
```diff
@@ -6,7 +6,6 @@
  *)
 
 open! IStd
-open Core
```
jvillard: Why no `Core_kernel` here?

infer/src/base/Process.ml
```diff
@@ -36,6 +36,7 @@ let create_process_and_wait ~prog ~args =
 
 
 let pipeline ~producer_prog ~producer_args ~consumer_prog ~consumer_args =
+  let open Core in (* fork *)
```
jvillard: Maybe if `Core_kernel` is the new god then we want to be more cautious about `open`ing `Core`? We could just use `let module Unix = Core.Unix in` here.

infer/src/base/Process.mli
```diff
@@ -6,6 +6,7 @@
  *)
 
 open! IStd
+open Core (* exit_or_signal *)
```
jvillard: Likewise here, we could just remove this `open` and be explicit when we use `Core`.

infer/src/base/ProcessPool.ml
```diff
@@ -8,6 +8,7 @@
 open! IStd
 module F = Format
 module L = Logging
+open Core (* select_fds, wait, signals, etc *)
```
jvillard: Same remark about opening `Core`.

infer/src/istd/IStd.ml
```diff
@@ -5,10 +5,374 @@
...
+    try
+      if stat_check_exn ~follow_symlinks f path then
+        `Yes
+      else
+        `No
+    with
+    | Unix.Unix_error ((Unix.EACCES|Unix.ELOOP), _, _)   -> `Unknown
+  ;;
+
+  let file_exists     = stat_check     (fun _ -> true)
```
jvillard: Can you run ocamlformat? :) `make fmt` should sort it out for you.

infer/src/istd/IStd.ml
```diff
@@ -5,10 +5,374 @@
  * LICENSE file in the root directory of this source tree.
  *)
 
-include Core_kernel
 
 [@@@warning "-32"]
 
+module Sys = struct
```
jvillard: Is that code copied from somewhere else? If so the copyright notices probably need changing. Maybe best to put them in their own file too (either all the modules in one file or one per module, not sure which is best).

infer/src/bufferoverrun/bufferOverrunDomainRelation.ml
yakobowski: The entire `open` was unused (as it is implicitly done by `open IStd`), so I removed it.

infer/src/istd/IStd.ml
yakobowski: Sure :smile:  Do you want me to re-indent the existing commit and push -f, or add another commit?

infer/src/istd/IStd.ml
yakobowski: Most of the code here is copied from `Core` itself indeed, except for the completely trivial stubs (typically labelling functions). In any case, creating a separate function was on my TODO list, and I will adapt the copyright accordingly.

infer/src/istd/IStd.ml
jvillard: push -f sounds good.

infer/src/base/Process.ml
yakobowski: `Process` and `ProcessPool` are the two files on which I'm a bit stuck, as it is highly non-trivial (I believe) to rewrite them not to use Core. Any suggestion would be appreciated. In particular, I think I remember @mbouaziz mentioning that those files were not needed for the Java backend (or maybe I misunderstood). If so, one possibility would be to isolate a reduced part of Infer in which processes are not used, perhaps available through a config switch?

jvillard: They are used to parallelise the analysis. One option could be to have a configure switch to enable/disable them and forego parallelism on Windows for the time being. This involves moving these files to their own directory and then play the same trick as for clang/ vs clang_stubs/ and java/ vs java_stubs/: if the switch is true then pick one if not the other, but both directories contain the same toplevel interface to the outside world (the _stubs one being just a dummy implementation–in that case a sequential one).

Longer term (hopefully just a few months), we plan to work on using the Hack infrastructure for parallelism instead of this ad-hoc "ParMap"-like implementation, which I believe is compatible with Windows.

infer/src/istd/IStd.ml
yakobowski: Done, in a new file `Core_compat`.

infer/src/istd/IStd.ml
yakobowski: Done

yakobowski:
PR updated, it is ready for a second round of review. `Core` as a dependency is dropped entirely. The important C functions have been stubbed, so no functionality should be lost (barring bugs, of course).

There remains one problem, causing two tests to fail, and making `infer.bc` and `infertop.bc` hard to use. The `core_stubs.c` file that I have added force the creation of a dll `dllInferStdlib_stubs.so`, which is not found by default. The only solution I found was to force a `dune build -p InferStdlib`, which creates symlinks from the built dll to the installation directories under `_build`. I doubt this is the best solution, and I'm not sure where I should insert this command anyway. Any help would be appreciated. BTW, this is probably related to https://github.com/ocaml/dune/issues/108

FWIW, the current state of the Windows port is in https://github.com/yakobowski/infer/tree/win The first 4 commits are ready, the next ones remain to be refined.

infer/src/base/Process.ml
yakobowski: The non-parallel version of `ProcessPool` is in f395c1bae7f5345f1fe820b421f06b33738e94e6. The build trick machinery I used is a bit different, but it works. Your alternative was not playing nice for a single file in the middle of a directory/library.

yakobowski:
TODOs after a live discussion:
- [ ] inline as most of `Core_compat` as it makes sense, as `Core` is going to be dropped as a dependency on all architectures
- [ ] remove the function that uses `flock` to kill the dependency

For the "Windows" part of the migration, which has not been submitted yet:
- [ ] for `realpath`, assuming it does nothing (as is the case under Windows), check that the binaries that are not `infer` (i.e. `infer-capture` and co) still work
- [ ] check that installation works when `ln` is `cp`
- [ ] see whether it would be more efficient to rewrite `Tasks` instead of `ProcessPool`

yakobowski:
See https://github.com/facebook/infer/pull/1066 for the removal of `flock`.

yakobowski:
The bug with the bytecode executables is now fixed, I will let you judge whether other parts of `Core_compat` can/should be inlined. It stopped when the Core API had clear benefits. (Less ambiguity thanks to labelled arguments, better return types, etc.)

jvillard:
This is shaping up nicely! I'm not so scared about the amount of copied stuff in Core_compat.ml anymore since it seems small enough. Can I ask the following:
- put Core_compat.ml and its stubs in their own folder, optionally their own dune library as it shouldn't be much more difficult to do, eg infer/third-party/core-compat/Core.ml?
- Carve out parts of Core_compat.ml that you wrote yourself and put them in a new file infer/src/istd/IUnix.ml instead, to be consistent with how we add functionalities to other standard modules (eg IList.ml, IOption.ml, ...)
- add .mli when possible
- what's up with the remaining TODOs?

yakobowski:
Sure! Will do ASAP.

Regarding the TODOs:
- the one for sigkill instead of sigint is more for you, to confirm that you are ok with this approach
- the one on ``My_group | `Group _` can probably be removed by forbidding those variants altogether. I will do this
- the one for `putenv` is minor, but again more of a question for you: do you think you use the presence of a (possibly empty) environment variable somewhere?
- the one for `sexp_or_error` needs to be cleaned: either I managed to get an sexp, or I remove the TODO, it is completely minor.

XVilka:
There are some conflicts now, so should be rebased. Was there any progress since?

XVilka:
Ping?

yakobowski:
Sorry, very busy summer. I will rebase the PR on work on this this week.

XVilka:
How the rebase did go?

andrewvaughanj:
I guess that this branch is in relation to https://pldi20.sigplan.org/details/infer-2020-talks/3/-Infer-ring-things-on-Ada-code-Invited-Talk

Would be great if we could see some movement on the Windows build side of things :)

@yakobowski do you know if AdaCore are planning on releasing modifications to Infer to use `libadalang`?

yakobowski:
@andrewvaughanj there is a link indeed ;)

Without spoiling too much my own talk, we (AdaCore) definitely intend to do what's needed to get this merged. In fact, I'm working right now on the "sequel" of this, namely to get parallelism under Windows. Unfortunately, for all intents and purposes the version of Infer we used is a fork which dates back to the origin of this branch. We have a lot of work to do, first to rebase everything, then to contribute back what's meaningful.

Are you personally interested in a Windows version BTW?

andrewvaughanj:
> Are you personally interested in a Windows version BTW?
>

To a degree, yes; as part of my employment, we considered building something on top of Infer, but given the Windows port wasn't "ready" (and that wasn't our focus), we took a different approach.

For sure though, I think the Windows port has a lot of interest to industrial users of Infer.

For context, I did the Windows port of Boolector (which we released back as MIT), so this kind of work is interesting to me.

jvillard:
I've converted this to a draft pending updates. Not sure what this means though, let me know if that was a bad idea.

yakobowski:
FYI, I managed to rebase this PR on last week's Infer. Are you still ok on the principle of it (i.e. replacing Core by Core_kernel)? Also, I assume that the remarks I did not address last year still stand? (https://github.com/facebook/infer/pull/1042#issuecomment-474374085)

yakobowski:
Closing, as this one is hopelessly out-of-date. I will start sending changes piecewise momentarily.

PR #1060, opened 2019-02-08
fix installation of alias binaries

yakobowski:
The former Makefile does not work when $(LN) is cp, because
it tries to copy a binary that has not been installed yet


yakobowski:
This PR does not take into account bytecode compilation, I will update it shortly.

yakobowski:
PR updated. Under Windows, `make byte` will remove `infer.exe`, but this is more or less unavoidable.

infer/src/Makefile
```diff
@@ -122,15 +122,20 @@ doc: $(SRC_BUILD_COMMON) $(MAKEFILE_LIST)
 
 INFER_BIN_ALIASES = $(foreach alias,$(INFER_COMMANDS),$(BIN_DIR)/$(alias))
 
-$(INFER_BIN_ALIASES): Makefile
-#	make sure the bin directory exists
-	$(QUIET)$(MKDIR_P) $(@D)
+$(INFER_BIN_ALIASES): Makefile $(BIN_DIR)/$(INFER_MAIN)
+	$(REMOVE) $(@D)/$(@F)
```
jvillard: nit: Isn't `$(@D)/$(@F)` the same as `$@`?

infer/src/Makefile
```diff
@@ -122,15 +122,20 @@ doc: $(SRC_BUILD_COMMON) $(MAKEFILE_LIST)
...
 	$(QUIET)cd $(@D) && $(LN_S) -f infer $(@F)
 	$(QUIET)touch $@
 
-$(INFER_BIN).exe: $(INFER_BUILD_DIR)/$(INFER_MAIN).exe $(INFER_BIN_ALIASES)
+$(INFER_BIN).exe: $(INFER_BUILD_DIR)/$(INFER_MAIN).exe
+	$(QUIET)$(MKDIR_P) $(BIN_DIR)
+ifeq ($(OS), Windows_NT)
```
jvillard: What's `OS`?

infer/src/Makefile
yakobowski: Indeed. I will simplify.

infer/src/Makefile
yakobowski: An environment variable that is frequently used to detect the OS on which the build executes. See e.g. https://stackoverflow.com/questions/714100/os-detecting-makefile

yakobowski: I removed the space after the comma just in case.

jvillard: I think it would be better to put this detection in configure.ac and export a variable in Makefile.autoconf.in, similarly to other build-time configuration. The internet suggests using `AC_CANONICAL_HOST`: https://stackoverflow.com/a/38899152

yakobowski: As discussed: using `AC_CANONICAL_HOST` forces the inclusion `config.sub` and `config.guess` in the tree, and we do not want that. So I exported `OS` in the `configure` under a meaningful name instead.

PR #1067, opened 2019-02-25
Fix minor typo

yakobowski:


yakobowski:
The code was obviously working without the conditional, so it might be simpler to drop it entirely.

jvillard:
Oops :) I guess nothing uses these "extra cflags" anymore, probably better to delete altogether I agree.

PR #1100, opened 2019-05-14
Point to the latest release

i-ky:
Newcomers may not be aware which is the latest Infer version and will likely download version mentioned in _Getting started_ guide, so it should point to the latest release. Unfortunately, it is not possible to just point to the [latest](https://github.com/facebook/infer/releases/latest) release, because asset name currently contains version.


jvillard:
Thanks!

PR #1102, opened 2019-05-21
Improvements in docker container generation

jesmg:
Using "opam init" in non interactive mode, in order to avoid compilation failures. Otherwise, It will wait for user input until a timeout, that will cause the compilation process to fail due to an OPAM error.

Also the Dockerfile has been modified in order to use [multistage builds](https://docs.docker.com/develop/develop-images/multistage-build/), obtaining a cleaner Docker image.

### DIFF
--- a/docker/master/Dockerfile
+++ b/docker/master/Dockerfile
@@ -22,12 +22,12 @@ RUN apt-get update && \
       make \
       openjdk-8-jdk-headless \
       patch \
+      patchelf \
       pkg-config \
       python2.7 \
       unzip \
       xz-utils \
-      zlib1g-dev \
-      patchelf && \
+      zlib1g-dev && \
     rm -rf /var/lib/apt/lists/*
 
 # Some scripts in facebook-clang-plugins assume "python" is available
@@ -39,7 +39,7 @@ RUN curl -sL https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x8
 
 # Disable sandboxing
 # Without this opam fails to compile OCaml for some reason. We don't need sandboxing inside a Docker container anyway.
-RUN opam init --reinit --bare --disable-sandboxing -ya
+RUN opam init --reinit --bare --disable-sandboxing --yes --auto-setup
 
 # Download the latest Infer master
 RUN cd / && \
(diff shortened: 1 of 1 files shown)


mbouaziz:
Thanks!

docker/master/Dockerfile
```diff
@@ -26,7 +26,8 @@ RUN apt-get update && \
       python2.7 \
       unzip \
       xz-utils \
-      zlib1g-dev && \
+      zlib1g-dev \
+      patchelf && \
```
jvillard: nit: please keep the list in alphabetical order

docker/master/Dockerfile
```diff
@@ -38,7 +39,7 @@ RUN curl -sL https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x8
 
 # Disable sandboxing
 # Without this opam fails to compile OCaml for some reason. We don't need sandboxing inside a Docker container anyway.
-RUN opam init --reinit --bare --disable-sandboxing
+RUN opam init --reinit --bare --disable-sandboxing -ya
```
jvillard: nit: other options are in --long-form, so best to keep it that way and use `--yes --auto-setup`.

docker/master/Dockerfile
```diff
@@ -54,20 +55,22 @@ RUN cd /infer && \
     ./configure && \
     ./facebook-clang-plugins/clang/setup.sh
 
-# Hackish for now: pull to get the latest version
-RUN cd /infer && git pull 
+# Generate a release
+RUN cd /infer && make install-with-libs BUILD_MODE=opt PATCHELF=patchelf DESTDIR="/infer-release" libdir_relative_to_bindir="../lib"
```
jvillard: nit: break long line?

docker/master/Dockerfile
```diff
@@ -54,20 +55,22 @@ RUN cd /infer && \
...
+
+FROM debian:stretch-slim AS executor 
+
+# Install python 2.7 since infer requires it to run
+RUN apt-get update && apt-get install --yes --no-install-recommends \
+    python2.7
```
jvillard: Why is `python2.7` needed here but none of the other dependencies from the compilator are? I'm surprised because some of them are also runtime dependencies, like `libgmp` and `libmpfr`. Or perhaps they're installed by default, only their `-dev` counterparts aren't?

docker/master/Dockerfile
jesmg: modified

docker/master/Dockerfile
jesmg: modified

docker/master/Dockerfile
jesmg: modified

docker/master/Dockerfile
jesmg: When generating the "release" version
 https://github.com/facebook/infer/blob/8ae6b1faf3d72d0bbafe352a3c9e15bc62e9d733/docker/master/Dockerfile#L59-L64
dependencies are moved to the "lib" folder (inside the installation path of Infer) and the binaries are patched in order to use them. But for any reason that is not made with python, so if you move the release version to another container without python2.7, it won't work.

PR #1119, opened 2019-06-30
Clarify the instructions for non-Homebrew downloads + Clarify instructions for gradle + Android Studio

ozmium:
Clarify the download instructions for direct downloads (without Homebrew)

* Indicate the file disk space used
* Add tar download instructions for MacOS, using curl and tar
* Add web links to the next sections, since they are not visible at all on the web site, when viewing in mobile view or tablet view.

----

Clarify instructions for gradle + Android Studio

* The user must cd into the folder that contains 'gradlew' first, then run the command, otherwise it will not work.
* Explain that `./gradlew clean` normally needs to run first, otherwise error.
* Explain that the gradle flag `--rerun-tasks` can be used instead of `clean`, which is more convenient to deleting built binaries.
* Use the more common `assembleDebug` command instead of `build`, to avoid building all flavors/variants in both debug and release mode.

### DIFF
--- a/_docs/00-getting-started.md
+++ b/_docs/00-getting-started.md
@@ -7,11 +7,12 @@ permalink: /docs/getting-started.html
 
 ## How to Download and Install Infer
 
-You can use Homebrew (Mac only), our binary releases, build infer from
-source, or use our Docker image. As of April 2019 (version 0.16.0),
-**it will require about 1.2 GB of disk space.**
+You can either install via Homebrew (Mac only), download our binary releases,
+build infer from source, or use our Docker image. As of April 2019 (version 0.16.0),
+**the binary releases will require about 1.2 GB of disk space.** Downloading the source
+code and building it will require a lot more than 1.2 GB, to compile clang for instance.
 
-# Use Homebrew (macOS only)
+### Use Homebrew (macOS only)
 
 On macOS, the simplest way is to use [Homebrew](http://brew.sh/). Type this into a terminal:
 
@@ -19,7 +20,7 @@ On macOS, the simplest way is to use [Homebrew](http://brew.sh/). Type this into
 brew install infer
 ```
 
-# Use a Direct Download from Github (Linux or macOS)
+### Use a Direct Download from Github (Linux or macOS)
 
 On Linux, or if you do not wish to use Homebrew on Mac, you can directly
 download our [latest binary
@@ -46,13 +47,16 @@ tar -xJf infer-osx-v0.16.0.tar.xz    # Unzip the compressed file. This will crea
 mv infer-osx-v0.16.0/ /Applications/ # Move the infer folder into /Applications
 rm infer-osx-v0.16.0.tar.xz          # Delete the original downloaded file
 
-# These next steps are optional, to put the infer binary in the PATH environment variable
+# These next steps are optional, to add the infer binary to the PATH environment variable
 
 # Open ~/.bash_profile in a text editor
 nano ~/.bash_profile
 
 # Add this line to the end of ".bash_profile". Then save the file with Ctrl+X, then Y, then Enter
 PATH=/Applications/infer-osx-v0.16.0/bin:$PATH
+
+# Alternatively, create a symbolic link to infer (instead of editing '.bash_profile')
+ln -s "/Applications/infer-osx-v0.16.0/bin/infer" /usr/local/bin/infer
 
 # Check that infer is on the PATH:
 type infer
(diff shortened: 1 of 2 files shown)


_docs/00-getting-started.md
```diff
@@ -5,23 +5,29 @@ layout: docs
 permalink: /docs/getting-started.html
 ---
 
-## Get Infer
+## How to Download and Install Infer
 
 You can use Homebrew (Mac only), our binary releases, build infer from
-source, or use our Docker image.
+source, or use our Docker image. As of April 2019 (version 0.16.0),
+**it will require about 1.2 GB of disk space.**
```
jvillard: Which of these 3 installations methods does 1.2 GB correspond to? I think that number should be moved to the relevant section of this document. Installing from source requires a lot more than 1.2 GB to compile clang for instance.

_docs/00-getting-started.md
```diff
@@ -5,23 +5,29 @@ layout: docs
...
 
-On Mac, the simplest way is to use [Homebrew](http://brew.sh/). Type this into a terminal:
+# Use Homebrew (macOS only)
```
jvillard: Should this be `###` instead of `#`?

## Get Infer
### Use Homebrew (macOS only)

_docs/00-getting-started.md
```diff
@@ -30,6 +36,28 @@ curl -sSL "https://github.com/facebook/infer/releases/download/v$VERSION/infer-l
 ln -s "/opt/infer-linux64-v$VERSION/bin/infer" /usr/local/bin/infer
 ```
 
+Use this to download the **Mac version** of infer to `/Applications` and optionally put
+it on the `PATH`. You should replace `0.16.0` with the [latest release version](https://github.com/facebook/infer/releases/latest):
+
+```sh
+curl -O -L "https://github.com/facebook/infer/releases/download/v0.16.0/infer-osx-v0.16.0.tar.xz"
```
jvillard: Why not copy the instructions for Linux, replacing /opt with /Applications? You could also re-use the symbolic link method to get infer in the `PATH` to avoid having to edit .bash_profile.

_docs/00-getting-started.md
```diff
@@ -47,3 +75,13 @@ sh run.sh
 ## Try Infer in your browser
 
 Try Infer on a small example on [Codeboard](https://codeboard.io/projects/11587?view=2.1-21.0-22.0).
+
+# Next Sections
+
+[//]: # (These links are required because they are completely hidden and inaccessible when the web page is viewed on a tablet or phone)
```
jvillard: Ah, sorry about that, I'll see if we can fix this. We have the links you added here in the meantime.

_docs/01-analyzing-apps-or-projects.md
```diff
@@ -58,11 +58,19 @@ infer run -- make -j 4
...
 
 
-### Gradle
+### Gradle / Android
+
+Normally, you would need to do a clean build first, otherwise infer will
+complain with error: `"Nothing to compile. Try cleaning the build first."`
+However, cleaning the build will delete all of your built binaries, which is
+inconvenient, so use the flag
+[`--rerun-tasks`](https://stackoverflow.com/questions/7289874/resetting-the-up-to-date-property-of-gradle-tasks)
```
jvillard: Good catch, thanks!

_docs/00-getting-started.md
ozmium: I can clarify this.

_docs/00-getting-started.md
ozmium: I don't know what the style guides are... I'll change it though.

_docs/00-getting-started.md
ozmium: I'll add the symbolic link option as well, but it has a disadvantage of creating another file in an obscure location. It's probably easy to forget to delete this file when uninstalling infer. But nevermind.

However, I can't reuse the Linux instructions because:

* The tar and curl flags used in the Linux example do not work on MacOS. I tried it, and it just started spitting out many MB of binary data to the Terminal, and I had to force quit the command.

* I consider it too complicated to combine all those commands into one line or one command. It's not clear what each part does, and it's hard to copy it as a long string. When it's broken up into separate commands, it's easier to understand what each part does. And the user can pause in between each step too.

* I consider the `"VERSION=0.XX.Y"` string substitution somewhat complicated for users who want to install the software manually. It's ok for something like an automated script, but not for manual typing. Another reason: The Terminal/shell will auto-complete filenames and paths when pressing Tab; but this won't happen if the filename contains a string substitution.

XVilka:
Should be this updated? Also related to https://github.com/facebook/infer/pull/1182

jvillard:
Some of the concerns addressed by this PR have been resolved since. The website has moved to the master branch under website/ too so this PR doesn't apply anymore. Closing.

PR #1177, opened 2019-11-01
Fix the ppx_deriving issue URL

XVilka:
Pretty trivial, but important in case GitHub will stop redirecting.


jvillard:
Good catch, thanks!

PR #1075, opened 2019-03-22
replace travis script to enable caching

jberdine:

PR #1181, opened 2019-11-05
Re-add OSX support for CI

Amar1729:
Re-add OSX support, and improve the general workflow of the current CI.

Edits made:

- (osx) use `homebrew` addon for system packages.

- (will result in safer builds on both osx and linux) Remove infer from opam switch before cache (keep builds more isolated, and prevent odd timeouts on archive packing on osx)
  - [`opam update --upgrade` fails](https://travis-ci.org/github/facebook/infer/builds/661635581?utm_source=github_status&utm_medium=notification) due to menhir/infer recompiling at the same time, while menhir is a dependency of infer (this happened in unrelated pr #1237 )
  - [archive timeout](https://travis-ci.org/github/facebook/infer/jobs/661141420?utm_medium=notification&utm_source=github_status) 

- (osx) do not run `opam depexts --upgrade` on macOS; if run, it will [unnecessarily call Homebrew](https://travis-ci.org/github/facebook/infer/jobs/662156953?utm_medium=notification&utm_source=github_status), resulting in a longer build

- (osx) Use aspcud on osx as an external solver, otherwise opam runs out of memory
  - [related build](https://travis-ci.org/github/facebook/infer/builds/660931619?utm_medium=notification&utm_source=github_status)
<details>
<pre>
opam(4068,0x7fffb59d9380) malloc: *** mach_vm_map(size=3039929774538752) failed (error code=3)
*** error: can't allocate region
*** set a breakpoint in malloc_error_break to debug
Fatal error: out of memory.
The command "opam update --upgrade" failed and exited with 2 during .
</pre>
</details>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,39 @@ addons:
   apt:
     packages:
     - libmpfr-dev
+  homebrew:
+    packages:
+      - aspcud
+      - automake
+      - gmp
+      - libtool
+      - lzlib
+      - mpfr
+      - pkg-config
+      - sqlite
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then ARCH="darwin"; else ARCH="linux"; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export PATH="/usr/bin:${PATH}"; fi
   - wget -O ${HOME}/opam https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x86_64-${ARCH}
   - chmod +x ${HOME}/opam
   - export PATH=${HOME}:${PATH}
   - export OPAMYES=1
   - export OPAMJOBS=2
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export OPAMEXTERNALSOLVER=$(which aspcud); fi
   - opam init --compiler=${OCAML_VERSION} --disable-sandboxing
   - eval $(opam env)
   - export OPAMVERBOSE=1
 install:
+  # ensure infer isn't installed in this switch, then deal with dependencies
+  - opam remove infer
   - opam update --upgrade
   - opam pin add --no-action infer .
-  - opam depext --update infer
+  - if [ "$TRAVIS_OS_NAME" != "osx" ]; then opam depext --update infer; fi
   - opam install --deps-only infer
 script:
   - opam install infer
+before_cache:
+  - opam remove infer
 os:
 - linux
+- osx


.travis.yml
```diff
@@ -6,12 +6,22 @@ cache:
...
 before_install:
-  - wget -O ${HOME}/opam https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x86_64-linux
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then ARCH="darwin"; else ARCH="linux"; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export PATH="/usr/bin:${PATH}"; fi
+  - wget -O ${HOME}/opam https://github.com/ocaml/opam/releases/download/2.0.3/opam-2.0.3-x86_64-${ARCH}
   - chmod +x ${HOME}/opam
   - export PATH=${HOME}:${PATH}
   - export OPAMYES=1
   - export OPAMJOBS=2
+  - rm -rf ${HOME}/.opam
```
jvillard: Note that there is hope that Travis caches the opam dir between runs (at least that was the motivation in #1075), so we do not want to force a deletion of the directory here (but maybe that was just for your test?).

Amar1729: Yup that was just for my test - I recall there being something odd about installation of some of the libraries against homebrew vs system paths when I was making the homebrew formula, so I just wanted to make sure that wasn't screwing anything up

Amar1729:
Hm I don't recognize the current error:

```
- [*ERROR**][50323] Capturing in javac mode...
- [*ERROR**][50323] Uncaught Internal Error: (Sys_error
- [*ERROR**][50323]   "models/com/facebook/infer/builtins/InferBuiltins.class]: No such file or directory")
- [*ERROR**][50323] Error backtrace:
- [*ERROR**][50323] Raised by primitive operation at file "stdlib.ml", line 390, characters 28-54
- [*ERROR**][50323] Called from file "stdlib.ml" (inlined), line 398, characters 2-47
- [*ERROR**][50323] Called from file "jFile.ml", line 174, characters 36-65
- [*ERROR**][50323] Called from file "java/jClasspath.ml", line 152, characters 16-57
- [*ERROR**][50323] Called from file "java/jMain.ml", line 141, characters 8-60
- [*ERROR**][50323] Called from file "integration/Javac.ml", line 112, characters 8-47
- [*ERROR**][50323] Called from file "integration/Driver.ml", line 314, characters 2-29
- [*ERROR**][50323] Called from file "base/Utils.ml", line 418, characters 16-20
- [*ERROR**][50323] Called from file "scuba/ScubaLogging.ml", line 66, characters 29-44
- [*ERROR**][50323] Called from file "infer.ml", line 21, characters 2-36
- [*ERROR**][50323] Called from file "base/Utils.ml", line 418, characters 16-20
- [*ERROR**][50323] Called from file "scuba/ScubaLogging.ml", line 66, characters 29-44
- [*ERROR**][50323] Called from file "infer.ml", line 148, characters 8-54
- [*ERROR**][50323]
- [*ERROR**][50323] Run the command again with `--keep-going` to try and ignore this error.
- [*ERROR**][50323] make[3]: *** [models/infer/report.json] Error 2
- [*ERROR**][50323] make[2]: *** [java] Error 2
- [*ERROR**][50323] make[1]: *** [infer_models] Error 2
- make: *** [infer] Error 2
[ERROR] The compilation of infer failed at "/usr/bin/make -j 1".
```

XVilka:
As a side note - you might want to add the Azure Pipelines support, it timeouts less than Travis CI. See on how to setup it:
- https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
- https://azure.microsoft.com/en-us/services/devops/pipelines/

Amar1729:
I checked out `git bisect` between [current master](https://github.com/facebook/infer/commit/087a5a9787351b28d3850dc9e4edb22a9b0a67bf) and [v0.17.0](https://github.com/facebook/infer/tree/0.17.0) to find where `infer` stops building on macOS. 

Based on success of `./build-infer.sh -y java`, the first bad commit (infer not compiling on macOS) seems to be [bf0d1d88613025037d66212ec300d3797b759d21](https://github.com/facebook/infer/commit/bf0d1d88613025037d66212ec300d3797b759d21). 

This is actually right after [the commit updating infer to ocaml 4.08](https://github.com/facebook/infer/commit/85659d885274ca129a0af2dbfd4c6800ae80e78d), but that commit still works on macOS.

Amar1729:
@XVilka I checked out Azure for this, but opam commands timed out (they didn't seem to succeed installing anything, with or without `--disable-sandboxing`)

Amar1729:
Huh. The [previous build for this PR](https://travis-ci.org/facebook/infer/builds/660786561?utm_medium=notification&utm_source=github_status) passed (which I thought was surprising, since infer didn't seem to build for me locally), but now the PR's build is failing due to `opam update` running out of memory.

jvillard:
That's weird: at that point in the script there should be no packages installed, which thus should take no work at at. But I guess there's a cache for opam that messes this up.

Alternatively we can also remove all these manual `opam` steps in favour of running just `./build-infer.sh`, which will use the opam.locked file (thus bypassing dependency resolution) and do all the opam magic.

jvillard:
`./build-infer.sh java`, actually, since otherwise this will end up compiling clang and that takes many forevers.

Amar1729:
Running `rm -rf ${HOME}/.opam` seemed to "fix" the problem, although it certainly shouldn't be necessary ...

As far as only building java: I don't know how worth it this would be, but have you considered setting up CI for the `facebook-clang-plugins` project also? That way this project's CI could just pull the relevant artifact from say github releases and have a much easier time building java and c analyzers.

Amar1729:
Huh - [from this build at least](https://travis-ci.org/github/facebook/infer/jobs/661141420?utm_medium=notification&utm_source=github_status), the timeout is from Travis trying to pack the cache. Wonder why that would be the case.

Amar1729:
Uncovered an issue with the .travis.yml: it ran `opam install infer`, but that installation would get saved in the cache because `~/.opam` was cached.

For the next CI job, `opam update --upgrade` would actually do the work of recompiling infer, since infer was already installed in the switch and changes had been detected. Then during the `script` stage, `opam install infer` had nothing to do since it had already been installed.

This PR will also add an `opam remove infer` to the `script` job so infer does NOT get saved in the opam cache between runs.

Amar1729:
The issue with a couple of the previous CI builds is that `opam --update upgrade` was recompiling three menhir libs and infer at the same time. 

For some reason, opam didn't realize that menhir is a dep of infer, so it wouldn't install menhir before trying to recompile infer ... which would then fail during the configure stage because menhir wasn't found.

For future CI builds, it may be safest to run `opam remove infer` both during `install` stage before doing anything with opam (in case of previous failed builds), and in the `script` stage after running `opam install infer` to ensure it doesn't get cached.

Amar1729:
Ok (knock on wood) but this seems to work now. I'm going to rebase these commits and then remove the WIP status for merge.

Amar1729:
Looks like that's about the best I can do - unfortunate that it makes the travis file a bit wonky with if statements here and there.

@jvillard , pretty sure this can be merged now for better CI on linux and osx.

jvillard:
This is great, thank you!

.travis.yml
```diff
@@ -9,22 +9,39 @@ addons:
...
   - export PATH=${HOME}:${PATH}
   - export OPAMYES=1
   - export OPAMJOBS=2
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export OPAMEXTERNALSOLVER=$(which aspcud); fi
   - opam init --compiler=${OCAML_VERSION} --disable-sandboxing
   - eval $(opam env)
   - export OPAMVERBOSE=1
 install:
+  # ensure infer isn't installed in this switch, then deal with dependencies
+  - opam remove infer
```
jvillard: This is sometimes crashing (#1240), maybe it should be `opam remove infer || true`?

Amar1729: Huh, I guess it should be. Weird that it never errored when I tested it. I'll make the update.

Issue #1129, opened 2019-07-28
Homebrew tap for infer


Amar1729:
Please make sure your issue is not addressed in the [FAQ](http://fbinfer.com/support.html#troubleshooting).

Please include the following information:
- [x] The version of infer from `infer --version`.
- [x] Your operating system and version, for example "Debian 9", "MacOS High Sierra", whether you are using Docker, etc.
- [ ] Which command you ran, for example `infer -- make`.
- [ ] The full output in a paste, for instance a [gist](https://gist.github.com/).
- [ ] If possible, a minimal example to reproduce your problem (for instance, some code where
    infer reports incorrectly, together with the way you run infer to reproduce the incorrect
    report).

Managed to make a homebrew formula and build bottles for infer 0.17.0 and facebook clang 8.0.0 here:
- https://github.com/Amar1729/homebrew-formulae/blob/69e164cd825387aaa9af28b70dd081202edf1961/infer.rb
- https://github.com/Amar1729/homebrew-formulae/blob/69e164cd825387aaa9af28b70dd081202edf1961/fb-clang.rb

You can install from my tap with `brew install infer`, which will install `infer 0.17.0`. I would recommend this organization hosting a tap itself rather than relying on mine, since I might easily miss releases of infer or fb clang. 

I wrote fb-clang as its own separate formula so that it can be prebuit as a bottle, speeding up compilation time of the infer build. It could also allow for adding a `head` directive to the formula so people can install with `brew install --HEAD infer`. Since `--HEAD` builds are built by the end user they would only have to wait ~15m instead of an hour and a half for compiling clang too.

This is mostly meant for reference (although it works). You could hosting it instead as, for example, `facebook/homebrew-infer`. Hosting as a tap rather than in brew-core allows you to not have to deal with figuring out how to patch opam to avoid an x11 dependency (and fix the other related problems that makes the build too complex for brew-core).

You hosting it could also help clear up the build process for clang and infer a bit, or just standardize the installation for people on darwin.  
- One possible improvement that may help is allowing an optional `--prefix` argument for clang's `./setup.sh` script to specify an install directory
  - current workaround to use a custom CLANG_PREFIX, since exporting CLANG_PREFIX by itself doesn't seem to work by itself:
```
    ENV["CLANG_PREFIX"] = Formula["fb-clang"].opt_prefix.to_s
    system "facebook-clang-plugins/clang/setup.sh", "-r"
[...]

PR #1182, opened 2019-11-06
Update macOS install with tap

Amar1729:
Add https://github.com/Amar1729/homebrew-infer as the way to install `infer` on Mojave.

#1129


jvillard:
Thanks for the tap! :tada:

PR #1206, opened 2019-12-17
Fix CI

waylon531:
The CI script complains because it's trying to build infer with an old version of ocaml, so to remedy this I bumped the ocaml version in CI. It looks like the CI passes with this change, I did some testing on my own fork of infer and got the CI job to pass in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: c
 sudo: required
 env:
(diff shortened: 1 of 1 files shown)


jvillard:
Thank you @waylon531, sorry we missed this when migrating to 4.08.

PR #1230, opened 2020-03-05
[website] Update docusaurus config

imbsky:


imbsky:
We have infer's Twitter account and GitHub repository, so we can replace the footer links.

jvillard:
Good catch, thanks.

imbsky:
Thanks for your review!

PR #1232, opened 2020-03-05
[website] change Algolia search id

yangshun:
Apologies! I tweaked the value but it was already imported into Phabricator and #1231 was merged without the change.


jvillard:
oops, thanks for the quick fix!

PR #1235, opened 2020-03-09
use apt addon in Travis

Amar1729:
clean up dependencies by using apt addon rather than `sudo apt`. This should make it easier to re-add osx support later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: c
-sudo: required
 env:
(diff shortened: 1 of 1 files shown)


Amar1729:
@jvillard this PR is partly to clean up the CI, partly to enable merge without conflict of pr #1181 - this way that PR will actually be able to run travis tests before merge instead of conflicting.

jvillard:
Looks good to me, thank you!

PR #1237, opened 2020-03-13
fix getting started link

rompic:
fixed broken link


rompic:
Same as https://github.com/facebook/infer/pull/1242

jvillard:
Hi @rompic, looks like you haven't signed the CLA so I'll merge #1242 instead if that's ok.

Issue #934, opened 2018-06-05
Support Python 3


mcandre:
Am I understanding this correctly, that Infer currently depends on Python 2.7? If so, it would be good to add support for Python 3, in order to run more easily in modern development environments, especially now that Python 2 will soon be dropped from receiving security updates.

jvillard:
Yes infer needs python2.7 for some of its build system integrations (buck, ant, gradle, and xcodebuild). If I understand correctly, Python2.7 will be supported until 1 Jan 2020.

The Python code could be rewritten in OCaml to simplify way infer works (currently OCaml calls into Python for the integrations above, which calls back into OCaml), or else be ported to Python 3 as we near 2020.

engineeringstuff032:
Integration

XVilka:
Note, that 2020 will happen in a couple months, so it might be a good time to revisit this.

XVilka:
I quickly skimmed through all python files in this repo - most of the code looks compatible with Python 3, you might just try it as is, there are couple errors, can't fix it myself until I am able to build it (see https://github.com/facebook/infer/issues/1194), but it's probably 15min maximum if you have a working build environment.

XVilka:
Would be nice if you quickly check these, since Python 2 is officially dead now, and many distributions already started to work on its removal actively.

XVilka:
Seems there is ongoing work to remove unnecessary Python code already:
- https://github.com/facebook/infer/commit/74e54575fac502f21bc0220703f57b15e62b82c8
- https://github.com/facebook/infer/commit/284c6fdb3b8533c1a1c6fcbddf0849f6bde231a2

jvillard:
@XVilka yes we are finally working on migrating the Python code in infer to OCaml, or at least the files that we call into from OCaml (which is most of it). We'll migrate whatever is left to python 3. I think that will be only [infer/tests/testlock.py](https://github.com/facebook/infer/blob/95807702fc9dfe0ae87b9d0d86f33ac13ab7a94e/infer/tests/testlock.py).

XVilka:
Can be closed since https://github.com/facebook/infer/commit/f867e14b89d3089bcab7f7d81e951af674e43183
Only Dockerfiles left to be updated. Maybe also switch to the Debian Buster as the base for these images?

PR #1247, opened 2020-04-01
Update base for Docker, remove python2

XVilka:
Last missing piece for [Python 2 removal](https://github.com/facebook/infer/issues/934) also updated Docker base for the latest Debian Stable - Buster. Also selected 4.08.1 for the Java build as suggested i https://github.com/facebook/infer/pull/1154#issuecomment-601144234

### DIFF
--- a/docker/master-java/Dockerfile
+++ b/docker/master-java/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch-slim
+FROM debian:buster-slim
 
 LABEL maintainer "Infer team"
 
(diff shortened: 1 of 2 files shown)


docker/master/Dockerfile
```diff
@@ -63,9 +60,9 @@ RUN cd /infer && \
 
 FROM debian:stretch-slim AS executor
 
-# Install python 2.7 since infer requires it to run
+# Install python 3.7 since infer Clang plugins requires it to run
```
jvillard, line 63: To be pedantic, the plugins only need python for the build I think.

docker/master-java/Dockerfile
```diff
@@ -18,19 +18,16 @@ RUN apt-get update && \
       libmpfr-dev \
       libsqlite3-dev \
       make \
-      openjdk-8-jdk-headless \
+      openjdk-11-jdk-headless \
```
jvillard, line 21: Did you test if infer works with java11? Eg `infer -- javac examples/Hello.java`.

jvillard:
Looks good, thanks!

XVilka:
There seems to be a bug with Debian/Ubuntu Python:
- https://bugs.launchpad.net/ubuntu/+source/apt-btrfs-snapshot/+bug/1763923
- https://github.com/nodejs/node/issues/30189
So it needs python3-distutils installation, also infer needs sqlite3.

docker/master-java/Dockerfile
XVilka: Yes, it works.

docker/master/Dockerfile
XVilka: Changed this line

docker/master/Dockerfile
```diff
@@ -63,9 +61,7 @@ RUN cd /infer && \
 
 FROM debian:stretch-slim AS executor
```
skcho, line 62:
```suggestion
FROM debian:buster-slim AS executor
```

Issue #1252, opened 2020-04-24
How do i view the issues in HTML report? for maven projects. Currently using Infer version v0.17.0-581af48 running on ubuntu. 


PradeepShetty14:
Please make sure your issue is not addressed in the [FAQ](http://fbinfer.com/support.html#troubleshooting).

Please include the following information:
- [ ] The version of infer from `infer --version`.
- [ ] Your operating system and version, for example "Debian 9", "MacOS High Sierra", whether you are using Docker, etc.
- [ ] Which command you ran, for example `infer -- make`.
- [ ] The full output in a paste, for instance a [gist](https://gist.github.com/).
- [ ] If possible, a minimal example to reproduce your problem (for instance, some code where
    infer reports incorrectly, together with the way you run infer to reproduce the incorrect
    report).

ezgicicek:
Hi,
You can run infer with debug option (`--debug` or shortcut `-g`) and then proceed to `<infer-out>/captured/<file-name>.html` to see the html output. 

I am not exactly sure for maven. I suppose you would need to look at the infer-out directory for your maven project.

PradeepShetty14:
Hi,
If we use the --debug option then it provides a HTML file which contains the source file.
I don't need that. I need a summary of the entire projects and issues. 
https://fbinfer.com/docs/advanced-features/

ezgicicek:
Ok, quoting from [https://fbinfer.com/docs/advanced-features/](https://fbinfer.com/docs/advanced-features/)

> this HTML file contains the source file. In each line of the file there are links to the nodes of the control flow graph that correspond to that line of code. So one can see what the translation looks like. Moreover, when you click on those links you can see details of the symbolic execution of that particular node. If the option --no_test is also passed to infer, then the page pointed to from the nodes contains the printout of the whole symbolic execution.

if you had the following simple example:

```
int null_deref_bad(){
  int* x = nullptr;
  return *x;
}
```

The html would look like the following: 

![Screenshot 2020-04-24 at 11 01 16](https://user-images.githubusercontent.com/6132408/80200726-f7082800-861a-11ea-9cbc-276bd7a699ce.png)

where you can click on the name of the function and see the Infer  "summary" (i.e. the abstract analysis result for the function) along with all the intermediate results. Moreover, it also inlines the issue infer found.

>  I need a summary of the entire projects and issues.

What exactly do you mean by the "summary" here? Do you wanna see all the issues found in the project? 

If so, refer to
[...]

PR #1253, opened 2020-04-24
Removing references to the previously removed (and now incorrect) '--no-test'

andrewvaughanj:
Signed-off-by: Andrew V. Jones <andrew.jones@vector.com>

## Issue

Commit [068622c](https://github.com/facebook/infer/commit/068622c) renamed the option `test` to be `only-cheap-debug` and also flipped the default value. Therefore, what used to be `--no-test` (i.e., you _do_ want expensive debug information) is now `--no-only-cheap-debug`.

However, [`01-advanced-features.md`](https://github.com/facebook/infer/blob/master/website/docs/01-advanced-features.md) listed the option as `no_test` (with an underscore rather than a hyphen), which I guess got missed when renaming the option/changing the default.

The suggestion to use `--no_test` was recently suggested in [#1252](https://github.com/facebook/infer/issues/1252#issuecomment-618922148), so we should probably remove this, otherwise you get (even if you use the "correct" syntax for an older Infer):

```
$ infer --no-test
/path/to/infer/bin/infer: unknown option '--no-test'.
```

## Description of changes

Given that `only-cheap-debug` now has a default of _false_ (i.e., if you request `-g`/`--debug`, you automatically get `--no-only-cheap-debug`), making the suggestion to users to turn _off_ "no expensive debugging" is not required.

Consequently, this PR removes the recommendation to pass `--no-test`/`--no-only-cheap-debug` in the documentation.

### DIFF
--- a/website/docs/01-advanced-features.md
+++ b/website/docs/01-advanced-features.md
@@ -66,9 +66,7 @@ Moreover, we obtain an HTML page for each captured file inside
 the file there are links to the nodes of the control flow graph that correspond
 to that line of code. So one can see what the translation looks like. Moreover,
 when you click on those links you can see details of the symbolic execution of
-that particular node. If the option `--no_test` is also passed to `infer`, then
-the page pointed to from the nodes contains the printout of the whole symbolic
-execution.
+that particular node.
 
 ## Print the specs
 


wiji2523:
So how can unblock M

wiji2523:
If remove referracne which things we can use ?

wiji2523:
> If remove referecnes which things can use ?

andrewvaughanj:
I get the feeling that @wiji2523 might not be a genuine GH account ...

jvillard:
thanks @andrewvaughanj!

PR #1261, opened 2020-05-18
fix link to FAQ in issue template

tyrasd:
The current [link](http://fbinfer.com/support.html#troubleshooting) results in a 404 "page not found" error. Also updated to https since the website redirects to https anyway.


jvillard:
Good catch, thank you @tyrasd.

PR #1283, opened 2020-06-15
When building clang, add support for 'ninja' and for only using one process for linking

andrewvaughanj:
## Issue

There are a number of times when users building `facebook-clang-plugins` hit memory limitations, typically due to the OOM killer and when linking.

In https://github.com/facebook/facebook-clang-plugins/pull/21, I implemented support for building `facebook-clang-plugins` using both Ninja and to only use one process when linking.

## Resolution

This PR exposes the optional flags added in https://github.com/facebook/facebook-clang-plugins/pull/21 to `build-infer.sh`, such that you can build `facebook-clang-plugins` with either just `ninja` or `ninja` + only one process for linking.

As part of this, it updates the `facebook-clang-plugins` submodule.


andrewvaughanj:
Hmm, it occurs to me that we now have a `-y` (for `yes`) and `-n` (for `ninja`) -- is this confusing? Should we only have these options as "long" options?

INSTALL.md
```diff
@@ -19,6 +19,8 @@ is required to compile everything from source.
 - pkg-config
 - Java (only needed for the Java analysis)
 - cmake (only needed for the C/Objective-C analysis)
+- ninja-build (optional, if you wish to use sequential linking when building
+  the C/Objective-C analysis)
```
jvillard, line 23: Simpler? Also the tool is called "ninja" not "ninja-build".

```
- the ninja build system (optional, only used when building clang from source)
```

build-infer.sh
```diff
@@ -31,10 +31,12 @@ function usage() {
   echo "   java     build Java analyzer"
   echo
   echo " options:"
-  echo "   -h,--help             show this message"
   echo "   --no-opam-lock        do not use the opam.locked file and let opam resolve dependencies"
   echo "   --only-setup-opam     initialize opam, install the opam dependencies of infer, and exit"
   echo "   --user-opam-switch    use the current opam switch to install infer (default: $INFER_OPAM_DEFAULT_SWITCH)"
+  echo "   -h,--help             show this message"
+  echo "   -n,--ninja            use ninja for building clang (clang only)"
+  echo "   -s,--sequential-link  only use one process for linking (ninja only, clang only)"
```
jvillard, line 39: I think we could just have `-n,--ninja` and have it automatically enable sequential linking. This seems the only sane way to build clang to me so I see no good use case for having `-s` too.

Please keep the list alphabetically sorted (except it ignores "no-" prefixes).

Makefile
```diff
@@ -367,7 +367,7 @@ clang_setup:
 	    echo '$(TERM_INFO)(TIP: you can also force a clang rebuild by removing $(FCP_DIR)/clang/installed.version)$(TERM_RESET)' >&2; \
 	    echo >&2 ; \
 	  fi; \
-	  $(FCP_DIR)/clang/setup.sh; \
+	  $(FCP_DIR)/clang/setup.sh $(FCP_CLANG_USE_NINJA) $(FCP_CLANG_SEQUENTIAL_LINK); \
```
jvillard, line 370: Makefile variables that are used during the build like that are usually filled in by `./configure`. It would even allow us to auto-detect if `ninja` is available and then we won't need to add any options to `./build-infer.sh`.

andrewvaughanj:
> Thanks! I made some comments in-line but repeating the main takeaway here: I think it would be better to have "configure.ac" detect if "ninja" is available on the system by adding a `NINJA` variable to Makefile.autoconf.in and the appropriate logic in configure.ac (with just `AC_CHECK_TOOL([NINJA], [ninja], [no])` I think).
> 
> Then in the main Makefile, if `NINJA` is not `no`, passed the `--ninja` and `--sequential-link` options to ./clang/setup.sh.

Okay, this seems reasonable. Let me take a look at this and come back to you.

I think the other comments will resolve themselves (mostly) in making this change 😃

INSTALL.md
XVilka: Maybe change "cmake" to "CMake" too then?

andrewvaughanj: Done!

Makefile
andrewvaughanj: Done!

build-infer.sh
andrewvaughanj: No longer relevant.

andrewvaughanj:
@jvillard I think this captured it all; your solution was *a lot* better than what I had originally implemented!

Makefile
```diff
@@ -301,6 +301,10 @@ ifeq ($(BUILD_C_ANALYZERS),yes)
 byte src_build src_build_common test_build: clang_plugin
 endif
 
+ifneq ($(NINJA),no)
+FCP_COMPILE_ARGS = "--ninja --sequential-link"
```
jvillard, line 305: I think these quotes make this look like a single argument to the fcp setup script and should be removed. Does it work for you? It complains about the arguments on my machine.

jvillard: I'll just amend the commit, no worries.

andrewvaughanj: Yep, it works for me -- what platform? I tested this on Linux with GNU Make 4.2.1, so maybe it is that?

I have a `port` installed version of GNU Make on macOS (4.3); but, by default, macOS appears to have `3.81` (at least, that's what's in my `/usr/bin`).

jvillard: Fails on Debian bullseye and CentOS 7. Of course ninja has to be available to hit the bug. Removing the quotes fixes it.

andrewvaughanj: Fair enough ... want me to do it, or is it already done "from your end"?

jvillard: Already done.

jvillard:
Looks awesome, many thanks! These link-time OOMs have been a major pain in the ass, this should help a lot.

PR #1286, opened 2020-06-23
Update CI to build infer with c analyzers

Amar1729:
After merging https://github.com/facebook/facebook-clang-plugins/pull/20, we can now use Actions here to download a facebook-clang custom clang Release here and use it to build C Analyzers during CI, as well as Java Analyzers.

1. had to change `matrix.os: macos-latest` to `matrix.os: macOS-latest`. I guess that's what I typed for the fb clang build, and `curl`ing a Release asset is case-sensitive.

2. Add `jq` to installed packages on mac - we need it for parsing GitHub's API responses

3. Add the logic for fetching a release asset (we currently only look for the `"latest"` release) and downloading the properly-named clang based on OS. Then unpack it and update the `opam` files so we build with clang.

Note: I would have used this Action: https://github.com/dsaltares/fetch-gh-release-asset/ for fetching a Release asset, but unfortunately it seems to be in a fairly new state and it's a Docker Action, which means it doesn't support macOS. The logic in the fetch clang step is roughly based off of this action's logic.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -78,3 +78,44 @@ jobs:
       - run: opam install --deps-only infer
 
       - run: opam install infer
+
+      - name: Test infer
+        run: |
+          eval $(opam env)
+
+          echo -e "#include <stdio.h> \n int main() { int *s = NULL; *s = 42; return 0; } " > FailingTest.c
+          echo -e "#include <stdio.h> \n int main() { int *s = NULL; if (s != NULL) { *s = 42; } return 0; }" > PassingTest.c
+
+          # first command should exit with status 2
+          infer --fail-on-issue -P -- clang -c FailingTest.c || if [[ $? -ne 2 ]]; then exit 1; fi
+          infer --fail-on-issue -P -- clang -c PassingTest.c
+
+          echo 'class FailingTest {' >> FailingTest.java
+          echo '  String mayReturnNull(int i) {' >> FailingTest.java
+          echo '    if (i > 0) {' >> FailingTest.java
+          echo '      return "Hello, Infer!";' >> FailingTest.java
+          echo '    }' >> FailingTest.java
+          echo '    return null;' >> FailingTest.java
+          echo '  }' >> FailingTest.java
+          echo '  int mayCauseNPE() {' >> FailingTest.java
+          echo '    String s = mayReturnNull(0);' >> FailingTest.java
+          echo '    return s.length();' >> FailingTest.java
+          echo '  }' >> FailingTest.java
+          echo '}' >> FailingTest.java
+
+          echo "  class PassingTest {" >> PassingTest.java
+          echo "    String mayReturnNull(int i) {" >> PassingTest.java
+          echo "      if (i > 0) {" >> PassingTest.java
+          echo '        return "Hello, Infer!";' >> PassingTest.java
+          echo "      }" >> PassingTest.java
+          echo "      return null;" >> PassingTest.java
+          echo "    }" >> PassingTest.java
+          echo "    int mayCauseNPE() {" >> PassingTest.java
+          echo "      String s = mayReturnNull(0);" >> PassingTest.java
+          echo "      return s == null ? 0 : s.length();" >> PassingTest.java
+          echo "    }" >> PassingTest.java
+          echo "  }" >> PassingTest.java
+
+          # first command should exit with status 2
+          infer --fail-on-issue -P -- javac FailingTest.java || if [[ $? -ne 2 ]]; then exit 1; fi
+          infer --fail-on-issue -P -- javac PassingTest.java
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
+          - macOS-latest
           - ubuntu-latest
         ocaml-version:
           - 4.09.1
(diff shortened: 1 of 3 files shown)


Amar1729:
If https://github.com/dsaltares/fetch-gh-release-asset/ is updated to be a JS Action, this workflow should probably be updated to take out the custom `"Fetch clang Release"` step, (and the macOS installation of `jq` could be removed also).

Are there unit tests for infer that can be easily run after running `opam install infer`? Might help serve as an additional indicator for CI.

Amar1729:
Based on the output from the `opam install infer` step, I'm not entirely sure whether C analyzers are included. Is there an obvious point somewhere in the logs I'm missing that specifies which analyzers are enabled during a build?

Amar1729:
Ok, added some test commands that run at the end of the build. Currently, I'm getting an error that infer was built with C analyzers disabled, so there must be something I'm not doing right to enable them. Thoughts?

andrewvaughanj:
This is a total guess, but after extracting the clang build into place, have you tried running: `./facebook-clang-plugins/clang/setup.sh --only-record-install` to tell the Infer build process that we have a build of the plugins?

Amar1729:
@andrewvaughanj good idea, didn't work :/

My concern is during the `opam install infer` step, with the following output:
```
<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><>  🐫 
[NOTE] Ignoring uncommitted changes in /Users/runner/runners/2.263.0/work/infer/infer (`--working-dir' not active).
[infer.1.0.0] no changes from git+file:///Users/runner/runners/2.263.0/work/infer/infer#HEAD

The following actions will be performed:
  ∗ install infer 1.0.0*

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
∗ installed infer.1.0.0
Done.
# Run eval $(opam env) to update the current shell environment
```

Might need `opam install --working-dir` ? I'm unfamiliar with pinned packages that have dirty changes

andrewvaughanj:
Yeah, I think when you do your `sed` to remove `--disable-c-analyzers`, `opam install` ignores this local change, which is why it doesn't build Infer including the clang integration.

I tried this locally by swapping `--disable-c-analyzers` for `--disable-java-analyzers`, but running `opam install --working-dir .` didn't pick this up.

jvillard:
We could remove the `--disable-c-analyzers` from opam altogether. I don't think anyone uses the opam install except for the GitHub CI of infer.

jvillard:
Oh, the opam build wants to build clang again and then fails because there's not enough disk space.

Maybe we can give up on testing the build with opam for now, and just have the CI do `./build-infer.sh` (then `make test` if we are feeling ambitious, but can be done in another PR since this would be new functionality).

To be able to build from opam without rebuilding clang is going to be more complicated, especially now that opam is sandboxed by default. Some random thoughts about that:
- we could add a build option, eg to `./configure`, to be able to point the build system at another install of clang (and, eg, just copy it over to the correct place so everything else still works with no change needed) but that might not play well with the sandbox
- or, move the logic to download the corresponding facebook-clang-plugins release to the build scripts instead of the GitHub action. I'm not sure how easy that would be, these actions make it very easy to access the GitHub APIs. Again, the opam sandbox might not be happy about downloading random stuff from the internet during the build (and Homebrew will also be unhappy as that's against their guidelines). Probably a bad idea all around then.

Amar1729:
Seems like adding an optional build system to `./configure` would be the better way to solve the issue - I feel like adding more logic for downloading to the build scripts is unnecessary complexity.

For now, I'll switch the CI to use `build-infer.sh` to get it working.

.github/workflows/install.yml
```diff
@@ -35,18 +35,98 @@ jobs:
...
+          curl -J -L "$URL/releases/assets/$ASSET_ID" \
+              -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+              -H "Accept: application/octet-stream" \
+              -o "$CLANGFILE"
+
+      - run: tar -xzf clang-install-${{ matrix.os }}.tar.gz -C facebook-clang-plugins/clang/
+
+      - run: ./facebook-clang-plugins/clang/setup.sh --only-record-install
+
+      - run: ./build-infer.sh -y all
```
yakobowski, line 72: Replacing the `opam install infer` by this line is a major improvement IMO. Opam horribly truncates the build log, and it is almost impossible to understand why the build failed. This is much better!

Amar1729:
So looks like that works. I can keep the simple tests added here to ensure that infer under CI always builds with C and Java analyzers, or could remove them and add them in a different PR.

In other PRs, this workflow could probably be updated to run `make test` as well, and have better caching which would speed up builds a lot (the setup-opam action can be cached but I'm not sure how just yet, and I think `build-infer` should maybe take a --user-switch argument to use the cached opam packages? Not sure).

jvillard:
Thanks, this is awesome! Let's merge this.

Re: caching: the GitHub action already caches the opam stuff I think via the actions/cache action, although I don't think it's working as it seems to believe the cache was hit but empty:

```
Cache Size: ~0 MB (22 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/b51f080c-45c2-476c-bcdf-7a0aa6f7ba37/cache.tzst -P -C /home/runner/work/infer/infer
Cache restored from key: Linux-bd1d07754744236ae6e5e15ec80b133f56f03a080d50e233628ce5d82d46cc00
```

Passing `--user-opam-switch` to `./build-infer.sh` as you suggest would also help in the case where there's no cache hit by avoiding a recompilation of OCaml, I think.

Amar1729:
Looks like some internal tests failed. Anything I need to fix?

jvillard:
That was a spurious failure, landing now. Thanks again! :rocket:

PR #1287, opened 2020-07-01
typo in TopLifted from infer lab

GaloisNeko:
Pretty sure it should be `TopLifted`


jvillard:
Thank you @GaloisNeko!

PR #1290, opened 2020-07-15
add patchelf back to master/Dockerfile

tomtomjhj:
This was removed in D19272627 but not added back in D21998484.


jvillard:
Thank you @tomtomjhj!

PR #1295, opened 2020-07-22
fix(docs): Update README.md

iamdarshshah:
### Changes

- [x] Aligning header elements to the center. 
- [x] Added logo 

> This makes the docs look symmetrical and looks a bit catchy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,9 @@
-<p align="center">
-  <img src="website/static/img/logo.png" alt="logo" width="20%" />
-</p>
-<h1 align="center">
-  Infer
-</h1>
-<p align="center">
-  <a href="https://travis-ci.org/facebook/infer">
-    <img src="https://travis-ci.org/facebook/infer.svg?branch=master" alt="Build Status" />
-  </a>
-  <p align="center">
-    <a href="http://fbinfer.com/">Infer</a>is a static analysis tool for Java, C++, Objective-C, and C. Infer is written in <a href="https://ocaml.org/">OCaml</a>.
-  </p>
-</p>
+<img src="website/static/img/logo.png" alt="logo" width="15%" />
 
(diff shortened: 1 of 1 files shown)


jvillard:
I think it's better for maintainability to keep the markdown format here.

iamdarshshah:
Yes, agreed. But if we consider accessibility, then using `tags` will be beneficial. Apart from it, if you still want to go with markdown format, that's fine will update it to the original one and will add a logo to it.

Would that be good?

Thanks @jvillard 👍

jvillard:
Sorry, what do you mean by using tags?

Also note that the logo is available in the repo so you could use this instead of a link to fbinfer.com: https://github.com/facebook/infer/blob/master/website/static/img/logo.png

Not sure how snazzy we want to make this README though, there's already the website for the pretty-looking pages :)

iamdarshshah:
> Sorry, what do you mean by using tags?

By `tags`, I mean `HTML` tags which we can use for providing `Heading` to the document.

> Also note that the logo is available in the repo so you could use this instead of a link to fbinfer.com: https://github.com/facebook/infer/blob/master/website/static/img/logo.png

Sure, will update this

> Not sure how snazzy we want to make this README though, there's already the website for the pretty-looking pages :)

Totally agreed with this point. The website is really good but with that if we can make the GitHub documentation i.e., readme.md  for this project looks better then it will be good for the one who is checking this project for the first time and that too by visiting GitHub repo.

jvillard:
Can you elaborate why tags would be better for accessibility? In the final page the markdown is rendered as the appropriate html tags already, eg `# Infer` becomes `<h1>Infer</h1>`. I don't think html tags are desirable here.

iamdarshshah:
Agreed with your point. If you think html tags are not desirable here or better for maintainability we should avoid using html here, then will update it (this will make it aligned header elements to the left). But will keep the logo as of now.

Thanks @jvillard 👍

surajkumar8642:
I use this fast time.

PR #1311, opened 2020-09-24
Update website stuff

smorimoto:


smorimoto:
When I was looking at the Infer documentation, I noticed that the dark theme was still mostly incomplete. I think I will be able to take care of it in the near future, but for now it's better to temporarily disable it.

jvillard:
Thanks for your PR. I agree dark mode can use some love, but it seems mostly functional to me so maybe let's not kill it?

Can you remove the whitespace changes from your PR? I'm not sure what it actually changes and most of these files are automatically generated anyway.

smorimoto:
Of course. Then let's see if we can fix it right away.

jvillard:
@smorimoto could you please remove the other two commits so that only the dependencies upgrades remain?

smorimoto:
@jvillard Oh, I'm sorry, I forgot about this. I just updated everything and pushed the code that changed the format of the config file to the latest style.

jvillard:
Thank you, @smorimoto!

Issue #1083, opened 2019-04-10
DEAD STORE reports many harmless cases like int i = 0xFFFFFFFF; i = SomeFunc();


bmwvvvv:
infer version v0.15.0
in my case, 0xFFFF or 0xFFFFFFFF is a common initializer and there are a lot of this semantics in our code.  I would like to ignore those harmless.  but if I turn off liveness checker, I worry about missing some truely harmful bugs.  Does Infer provide some filtering option or method?

jvillard:
[This function](https://github.com/facebook/infer/blob/7e16aafdbaca9f6483a4002bd241d8085edf84ae/infer/src/checkers/liveness.ml#L219-L231) does what you want but only whitelists `x = 0`. I'd welcome a pull request to make the set of constants configurable by adding an option to [infer/src/base/Config.ml](https://github.com/facebook/infer/blob/7e16aafdbaca9f6483a4002bd241d8085edf84ae/infer/src/base/Config.ml).

bmwvvvv:
thanks! the whitelist can be added by an option. that's handful. I'll dig it.

XVilka:
Any updates on this one? @lucmaa

lucmaa:
sorry that there is none of it yet. @XVilka

2890148303:
I think it's so hard and I can't understand it completely.

jobnz:
I would be happy to do this, if no one else is working on it.

vickvey:
Hey i opened this as this was tagged with good first issue?? but how do i contribute into this and also i dont know what to contribute??!

PR #1340, opened 2020-11-02
Add a config option to whitelist common initializer values for liveness analysis

jobnz:
Implements #1083

### DIFF
--- a/infer/src/checkers/liveness.ml
+++ b/infer/src/checkers/liveness.ml
@@ -219,11 +219,23 @@ let checker {IntraproceduralAnalysis.proc_desc; err_log} =
   (* we don't want to report in harmless cases like int i = 0; if (...) { i = ... } else { i = ... }
      that create an intentional dead store as an attempt to imitate default value semantics.
      use dead stores to a "sentinel" value as a heuristic for ignoring this case *)
+  let is_whitelisted = function
+    | Some c ->
+        let rec helper c = function
+          | hd :: lst ->
+              if Int.(hd = c) then true else helper c lst
+          | [] ->
+              false
+        in
+        helper c Config.liveness_whitelist_constant
+    | None ->
+        false
+  in
   let rec is_sentinel_exp = function
     | Exp.Cast (_, e) ->
         is_sentinel_exp e
     | Exp.Const (Cint i) ->
-        IntLit.iszero i || IntLit.isnull i
+        IntLit.iszero i || IntLit.isnull i || is_whitelisted (IntLit.to_int i)
     | Exp.Const (Cfloat 0.0) ->
         true
     | _ ->
--- a/infer/src/checkers/liveness.ml
+++ b/infer/src/checkers/liveness.ml
@@ -215,7 +215,7 @@ module IntLitSet = Caml.Set.Make (IntLit)
 
 let whitelisted_constants =
   let int_lit_constants =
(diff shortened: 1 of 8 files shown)


infer/src/checkers/liveness.ml
```diff
@@ -219,11 +219,23 @@ let checker {IntraproceduralAnalysis.proc_desc; err_log} =
...
      use dead stores to a "sentinel" value as a heuristic for ignoring this case *)
+  let is_whitelisted = function
+    | Some c ->
+        let rec helper c = function
+          | hd :: lst ->
+              if Int.(hd = c) then true else helper c lst
+          | [] ->
+              false
+        in
+        helper c Config.liveness_whitelist_constant
```
jvillard, line 230: This looks like `List.mem ~equal:Int.equal Config.liveness_whitelist_constant c`.

infer/src/checkers/liveness.ml
```diff
@@ -219,11 +219,23 @@ let checker {IntraproceduralAnalysis.proc_desc; err_log} =
...
+    | None ->
+        false
+  in
   let rec is_sentinel_exp = function
     | Exp.Cast (_, e) ->
         is_sentinel_exp e
     | Exp.Const (Cint i) ->
-        IntLit.iszero i || IntLit.isnull i
+        IntLit.iszero i || IntLit.isnull i || is_whitelisted (IntLit.to_int i)
```
jvillard, line 238: I could imagine someone wanting to whitelist, eg MAX_INT, which wouldn't fit in an OCaml `int` but would fit in an `IntLit.t`. To handle this, the config option could be just a list of strings that is pre-processed somewhere in this module (liveness.ml) into a list (or better yet, a set) of `IntLit.t`.

jvillard (commented):
Looks useful to have, thank you! If you update your PR, could you also run `make manuals` to update the checked-in man pages?

infer/src/checkers/liveness.ml
```diff
@@ -211,6 +211,13 @@ let get_captured_by_ref_invariant_map proc_desc =
   let cfg = ProcCfg.Exceptional.from_pdesc proc_desc in
   CapturedByRefAnalyzer.exec_cfg cfg () ~initial:VarSet.empty
 
+module IntLitSet = Caml.Set.Make (IntLit)
+
+let whitelisted_constants =
+  let int_lit_constants =
+    List.map ~f:(fun el -> IntLit.of_string el) Config.liveness_whitelist_constant
```
jvillard, line 218: If the user passes bad input we'll crash here. Maybe worth getting a nice error message? E.g.

```
fun el -> try IntLit.of_string el with Invalid_argument -> L.die UserError "Ill-formed option  '%s' for --liveness-whitelist-constant: an integer was expected"
```

jvillard (commented):
Great stuff, thank you!

infer/src/checkers/liveness.ml
jobnz: Good point. Thank you for the helpful feedback!

jobnz:
Sorry, I'll undo the rebase

jvillard:
Thanks a lot @jobnz! I took the liberty of doing some minor changes on top of your PR and adding a test (see 606a3c95d9d9ca4b9ae7015b9c991c04e4d20ed3), hope that's ok with you.

jobnz:
Of course. Thanks again for the helpful reviews and comments!

PR #1341, opened 2020-11-04
Use `--depth` flag to `git clone` in Dockerfiles, to save space

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>


jvillard:
Thanks!

PR #1344, opened 2020-11-08
Fix Infer Installation Procedure

eankeen:
Similar to the `tar` command, append `sudo` to the `ln` command since it fails due to 'Permission Denied' errors on Linux


jvillard:
Thank you!

Issue #1002, opened 2018-10-01
Building Infer models error when executing build-infer.sh clang


harmim:
Command: `./build-infer.sh clang`
OS: macOS 10.13.6

```
$ ./build-infer.sh clang
...
[19:41:07][97874] Building Infer models...
[*ERROR**][97874] *** ERROR 'Building Infer models'
[*ERROR**][97874] *** command: ' /Library/Developer/CommandLineTools/usr/bin/make INTERACTIVE=1 infer_models'
[*ERROR**][97874] *** CWD: '/Users/harmim/Programs/infer'
[*ERROR**][97874] *** stdout:
[*ERROR**][97874] *** inner Building clang plugin
[*ERROR**][97874] *** inner command:  /Library/Developer/CommandLineTools/usr/bin/make INTERACTIVE=1 -C /Users/harmim/Programs/infer/facebook-clang-plugins/libtooling all CC="clang" CXX="clang++" CFLAGS="-g -O2" CXXFLAGS="-g -O2" CPP="clang -E" LDFLAGS="" LIBS="" LOCAL_CLANG=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install/bin/clang CLANG_PREFIX=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install CLANG_INCLUDES=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install/include
[*ERROR**][97874] *** inner CWD: /Users/harmim/Programs/infer
[*ERROR**][97874] make[2]: Nothing to be done for `all'.
[*ERROR**][97874] *** inner Building clang plugin OCaml interface
[*ERROR**][97874] *** inner command:  /Library/Developer/CommandLineTools/usr/bin/make INTERACTIVE=1 -C /Users/harmim/Programs/infer/facebook-clang-plugins/clang-ocaml all build/clang_ast_proj.ml build/clang_ast_proj.mli CC=clang CXX=clang++ CFLAGS="-g -O2" CXXFLAGS="-g -O2" CPP="clang -E" LDFLAGS="" LIBS="" LOCAL_CLANG=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install/bin/clang CLANG_PREFIX=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install CLANG_INCLUDES=/Users/harmim/Programs/infer/facebook-clang-plugins/clang/install/include
[...]

PR #1345, opened 2020-11-11
Fix of building Infer for Objective C

harmim:
- The SDK path is different for Objective C when command line tools for Xcode are installed without Xcode itself.
- Related issue https://github.com/facebook/infer/issues/1002.

I still have a problem with the mentioned issue. I think that it is possible that the SDK paths differs for Objective C on Mac OS when I have installed just command line tools for Xcode without Xcode itself. This change solves the issue for me. Is it a good solution? Could it be possibly added to Infer? Or is there any better way?


jvillard:
Thank you!

PR #1346, opened 2020-11-16
apache commons lang3 string utils model

fraimondi:
Small model for a couple of StringUtils functions


jvillard:
Thanks a lot @fraimondi, sorry for the delay.

PR #1372, opened 2021-01-02
Fix SIL to HIL conversion for Exp.Var inside Lfield and Lindex

kmh11:
When accessing a field or array offset of a pure variable (`Exp.Var`) that does not resolve to an access expression, `HilExp.of_sil` will create an extraneous dereference that causes `HilExp.get_typ` to fail. This pull request wraps variables that are the bases of Lfield or Lindex expressions with AddressOf before they're dereferenced (this is already done for Lvar inside `AccessExpression.of_pvar`) and adds a couple of unit tests that make sure it behaves as expected.

**More details on the bug:**

Given the following code:

```
if (!event_obj->dict)
```

and SIL:

```
n$6=_fun_gdb::ref_ptr<event_object,gdbpy_ref_policy<event_object>>::operator->(&event_obj:gdb::ref_ptr<event_object,gdbpy_ref_policy<event_object>>&) [line 38, column 8];
n$7=*n$6.dict:_object* [line 38, column 8];
PRUNE(!n$7, true); [line 38, column 8];
```

`operator->` has return type `event_object*`, but `n$6.dict` only has access to the type of the struct, `event_object`. `of_sil` [calls](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L567) `access_expr_of_lhs_exp` with that type, which [calls](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L498) `access_exprs_of_exp` (note that `add_deref` is always true). The Lfield case will then [recurse](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L469) to process the Exp.Var, and `AccessExpression.of_id` will return an `AccessPath.base` that is then [dereferenced](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L440). When resolving types, `get_typ` will find a non-pointer type wrapped by a `Dereference` and return [None](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L286). To fix this, we match what [of_pvar](https://github.com/facebook/infer/blob/9f98368e49d411941bb8e2a1e1839bed18cd625e/infer/src/absint/HilExp.ml#L295) does and wrap the base in an AddressOf, which is removed by the dereference.

### DIFF
--- a/infer/src/absint/HilExp.ml
+++ b/infer/src/absint/HilExp.ml
@@ -435,7 +435,7 @@ and access_exprs_of_exp ~include_array_indexes ~f_resolve_id ~add_deref exp0 typ
           in
           add_accesses access_expr' :: acc
       | None ->
-          let access_expr = AccessExpression.of_id id typ in
+          let access_expr = AccessExpression.address_of_base (AccessExpression.base_of_id id typ) in
           let access_expr' =
             if add_deref then AccessExpression.dereference access_expr else access_expr
           in
--- a/infer/src/inferunit.ml
+++ b/infer/src/inferunit.ml
@@ -34,6 +34,7 @@ let () =
         ; DifferentialTests.tests
         ; FileDiffTests.tests
         ; GradleTests.tests
+        ; HilExpTests.tests
         ; IListTests.tests
         ; JavaClassNameTests.tests
         ; JavaProfilerSamplesTest.tests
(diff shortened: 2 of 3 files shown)


jvillard:
Great catch, thank you @kmh11!

PR #1391, opened 2021-02-21
Set timeouts handlers only in biabduction mode

yakobowski:
Only register biabduction-style timeouts the first time the function
exe_timeout is called. This avoid getting timeouts in other
long-running analyses. (Especially on windows, where the wall clock is
used.)

### DIFF
--- a/infer/src/biabduction/Timeout.ml
+++ b/infer/src/biabduction/Timeout.ml
@@ -76,19 +76,23 @@ let timeout_action _ =
   raise (SymOp.Analysis_failure_exe FKtimeout)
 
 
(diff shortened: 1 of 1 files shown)


jvillard (commented):
Looks good to me, thanks! I guess --biabduction misbehaves on Windows, then, good to know. --pulse is set to replace it at some point.

yakobowski:
Right. We are not using bi-abduction ourselves, so I cannot really comment on the current state. But the timeout mechanisms is likely to introduce quirks given its current implementation, especially if the machine is (over)loaded.

PR #1392, opened 2021-02-21
Add support for -j0

yakobowski:
Use standard convention: 0 means the number of cpus

### DIFF
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -2997,7 +2997,12 @@ and javac_classes_out = !javac_classes_out
 
 and job_id = !job_id
 
-and jobs = Option.fold !max_jobs ~init:!jobs ~f:min
+and jobs =
+  let n = Option.fold !max_jobs ~init:!jobs ~f:min in
+  if n < 0 then L.(die UserError) "negative number of jobs specified (%d)" n
+  else if Int.equal n 0 then ncpu (* emulate standard convention *)
+  else n
+
 
 and join_cond = !join_cond
 


infer/src/base/Config.ml
```diff
@@ -2997,7 +2997,12 @@ and javac_classes_out = !javac_classes_out
 
 and job_id = !job_id
 
-and jobs = Option.fold !max_jobs ~init:!jobs ~f:min
+and jobs =
+  let n = Option.fold !max_jobs ~init:!jobs ~f:min in
+  if n < 0 then L.(die UserError) "negative number of jobs specified (%d)" n
+  else if Int.equal n 0 then ncpu (* emulate standard convention *)
```
jvillard, line 3003: This bypasses `--max-jobs`. Perhaps change the logic so that the 0 case is the same as the None case? Can you update the doc string (and run `make manuals`) to reflect the additional behaviour too?

infer/src/base/Config.ml
```diff
@@ -2997,7 +2997,12 @@ and javac_classes_out = !javac_classes_out
 
 and job_id = !job_id
 
-and jobs = Option.fold !max_jobs ~init:!jobs ~f:min
+and jobs =
+  let n = Option.fold !max_jobs ~init:!jobs ~f:min in
+  if n < 0 then L.(die UserError) "negative number of jobs specified (%d)" n
```
jvillard, line 3002: Thanks for that! nit: no need for the local open: `L.die UserError "..."`

yakobowski:
As you mentioned somewhere else (cannot find where right now), this is actually _not_ a standard convention. It is not accepted by `make`, and means `+\infty` for e.g. `ninja`. I also noticed we no longer use that feature internally, so it's simpler to close this PR. Sorry for the noise!

PR #1393, opened 2021-02-21
Fix compilation warning with gcc10

yakobowski:
gcc warnings are more strict starting from gcc10. Not having the const
qualifier triggers an error.


jvillard:
Thanks, that was showing up in my console sometimes!

PR #1408, opened 2021-03-12
Add support for correctly reading the 'TMPDIR' environment variable

andrewvaughanj:
Infer's [`install-sh`](https://github.com/facebook/infer/blob/master/install-sh) knows to "correctly" read from the `TMPDIR` environment variable:

- https://github.com/facebook/infer/blob/master/install-sh#L327

However, this was not supported by `clang/setup.sh`.

I haven't overlooked the existence of `CLANG_TMP_DIR`, but I had (initially) set `TMPDIR` hoping Infer's build system would honour it (it didn't) -- consequently, I thought it would be useful for future users if Infer honoured this.


jvillard:
Thank you for this.

PR #1423, opened 2021-04-01
Add Microsoft to List of Infer Users

matjin:
No functional changes


jvillard:
Thanks for your PR, this is great! The image has black text with a non-transparent background is not transparent, I can edit it or if you could upload another one (white text+logo, transparent background) that would be great.

PR #1424, opened 2021-04-01
[pulse] Fixed a bug in AbductiveDomain.leq operator.

mir597:
Summary: the mapping for computing 'leq' relation in isomorphic graphs was sometimes mixed with opposite mappings due to typos in the code.

### DIFF
--- a/infer/src/pulse/PulseAbductiveDomain.ml
+++ b/infer/src/pulse/PulseAbductiveDomain.ml
@@ -112,13 +112,13 @@ let leq ~lhs ~rhs =
      &&
      match
        BaseDomain.isograph_map BaseDomain.empty_mapping
-         ~lhs:(rhs.pre :> BaseDomain.t)
-         ~rhs:(lhs.pre :> BaseDomain.t)
+         ~lhs:(lhs.pre :> BaseDomain.t)
+         ~rhs:(rhs.pre :> BaseDomain.t)
      with
      | NotIsomorphic ->
          false
      | IsomorphicUpTo foot_mapping ->
-         BaseDomain.is_isograph (BaseDomain.invert_map foot_mapping)
+         BaseDomain.is_isograph foot_mapping
            ~lhs:(lhs.post :> BaseDomain.t)
            ~rhs:(rhs.post :> BaseDomain.t)
 
--- a/infer/src/pulse/PulseAbductiveDomain.ml
+++ b/infer/src/pulse/PulseAbductiveDomain.ml
@@ -112,13 +112,13 @@ let leq ~lhs ~rhs =
      &&
      match
        BaseDomain.isograph_map BaseDomain.empty_mapping
-         ~lhs:(lhs.pre :> BaseDomain.t)
-         ~rhs:(rhs.pre :> BaseDomain.t)
+         ~lhs:(rhs.pre :> BaseDomain.t)
+         ~rhs:(lhs.pre :> BaseDomain.t)
      with
      | NotIsomorphic ->
          false
      | IsomorphicUpTo foot_mapping ->
-         BaseDomain.is_isograph foot_mapping
+         BaseDomain.is_isograph (BaseDomain.invert_map foot_mapping)
            ~lhs:(lhs.post :> BaseDomain.t)
            ~rhs:(rhs.post :> BaseDomain.t)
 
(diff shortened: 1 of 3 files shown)


infer/src/pulse/PulseAbductiveDomain.ml
```diff
@@ -112,8 +112,8 @@ let leq ~lhs ~rhs =
      &&
      match
        BaseDomain.isograph_map BaseDomain.empty_mapping
-         ~lhs:(rhs.pre :> BaseDomain.t)
-         ~rhs:(lhs.pre :> BaseDomain.t)
+         ~lhs:(lhs.pre :> BaseDomain.t)
+         ~rhs:(rhs.pre :> BaseDomain.t)
```
jvillard, line 116: I think this was intentional but that your change is correct, and additionally the corresponding check for `post` below should be inverted. We want to establish `lhs |- rhs` (which the "Abstract Interpretation" framework will use to detect when loops have converged), i.e. `lhs` is more precise than `rhs`. In pulse, "more precise" means the graph for `lhs` has to be *larger* than the graph for the `rhs` (it has explored more of the state). But that's for post-conditions, which are in the positive direction. Pre-conditions need to reason by contraposition. So in the end: `lhs |- rhs <=> lhs.post ≥ rhs.post && lhs.pre ≤ rhs.pre`. That said I'm not sure there are examples where these directions matter. Were you able to find any?

infer/src/pulse/PulseBaseDomain.ml
```diff
@@ -125,7 +125,7 @@ module GraphComparison = struct
             IsomorphicUpTo mapping
         | Some _, None | None, Some _ ->
             NotIsomorphic
-        | Some (edges_rhs, attrs_rhs), Some (edges_lhs, attrs_lhs) ->
+        | Some (edges_lhs, attrs_lhs), Some (edges_rhs, attrs_rhs) ->
```
jvillard, line 128: This one was clearly a bug, great catch! I'm curious if you have a test case where it caused issues, we could add it to our tests if so but otherwise it's ok too.

jvillard (commented):
Great catch, thank you!

infer/src/pulse/PulseAbductiveDomain.ml
mir597: Ah, I understand the intention. This direction was correct. So it should invert the mapping for checking `post`. I will prepare new PR for this issue.
Still, since the result of the 'leq' operation of AbductiveDomain is a symmetric `lhs |- rhs <=> rhs |- lhs`, this patch would work well. 

I'm conducting a research on top of the pulse analyzer, and I found this bug while I'm debugging my work. So, I don't have a test case.

jvillard: Sorry, I think there was a misunderstanding. I think your previous version was correct, but to be *more* correct it should also swap lhs<->rhs in the post sub-graph check:
```
     | IsomorphicUpTo foot_mapping ->
         BaseDomain.is_isograph foot_mapping (* probably needs to be inverted mapping? *)
           ~lhs:(rhs.post :> BaseDomain.t)
           ~rhs:(lhs.post :> BaseDomain.t)
```

mir597: @jvillard  Please check the follow-up commit: a20a664
I think you are looking at a wrong commit. I already did it.
For pre, `isograph_map rhs.pre lhs.pre empty_map`
For post, `is_isograph lhs.post rhs.post (invert mapping)`

jvillard: Right, but I was wondering if it should rather be `isograph_map lhs.pre rhs.pre` and `is_isograph rhs.post lhs.post (invert mapping)`, but since isograph is symmetric it doesn't matter. I actually thought isograph was checking graph embedding, not equivalence, since there's a lhs and a rhs, but looking at it it's doing equivalence of the reachable parts. So maybe we shouldn't bother doing it one way and then the other and your first version was best after all!

mir597: I get the point. I thought that the meaning of 'is_isograph' should be `|-` relation between graphs, which is a dual of yours.
Thank you for your comments!

jvillard:
Thank you for this, Yoonseok!

PR #1431, opened 2021-04-19
Fix taskbar nesting for ondemand analyses

yakobowski:
On big analyses, we routinely observe analyses nesting levels of the
order of ~5k (i.e. 5342> on the interactive taskar). This call depth
is meaningless given the size of the codebase we analyze.

Instead, I believe this is due to the fact that Ondemand does not
properly snapshot/restore the nesting level when starting an analysis,
and more importantly when abandoning parts of it. This commit
fixes that oversight.


jvillard:
Thanks for the fix, @yakobowski!

PR #1434, opened 2021-04-23
Implement AtomicExpr translation

kmh11:
Implements the translation of most clang atomic builtins to SIL, including those used in `stdatomic.h`. It does not attempt to model the atomicity of the operations, since I don't know of any way Infer can represent that. I didn't bother implementing the rarely used min/max builtins, so they're left as `BuiltinDecl` calls.

This is my first major OCaml project, so any feedback is appreciated!

Also, CONTRIBUTING.md says to update the [facebook-clang-plugins](https://github.com/facebook/facebook-clang-plugins) submodule, but it doesn't seem to be a submodule anymore, and the code has diverged from that repo. Should I still make a PR over there?


jvillard:
This looks great, thank you! I'll review next week. The instructions are indeed outdated, the facebook-clang-plugins sub-module is now part of the infer repo itself. We mean to archive the facebook-clang-plugins repository.

infer/src/IR/Procdesc.mli
```diff
@@ -38,6 +38,8 @@ module Node : sig
   (** kind of statement node *)
   type stmt_nodekind =
     | AssertionFailure
+    | AtomicCompareExchangeBranch
+    | AtomicExpr of string
```
jvillard, line 42: A variant would be better than a string here. Or you could just leave out the argument to avoid copying the large type over to this file. This type is only used for debug anyway. The reason that a string is not ideal is that these end up in the CFGs written to disk and a string takes a lot more space than a variant (which is just an int).

jvillard (commented):
This looks in great shape already, thanks a lot for a very polished PR!

infer/tests/codetoanalyze/c/frontend/atomic_expr/arithmetic.c
```diff
@@ -0,0 +1,53 @@
+#include <stdatomic.h>
```
jvillard, line 1: Could you add copyright headers to each new file (except the generated .dot files)?

```
/*
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
```

infer/src/IR/Procdesc.mli
kmh11: I just removed the argument, which let me delete the massive `atomic_expr_to_string` function too.

kmh11:
Oops, I forgot to regenerate the dot files after changing the copyright headers because the line numbers changed... doing that now.

jvillard:
Thank you for your contribution, Kevin!

PR #1448, opened 2021-05-26
Fix matching double-quotes in json example

sangwoo-joh:
Hi.
Thanks for the great tool!

This is just a simple correction for string literals in the `Util` module.


jvillard:
Thank you @sangwoo-joh!

PR #1452, opened 2021-06-08
Forward type to implicit casts if part_of_explicit_cast

kmh11:
Currently, casts are not added in clang AST translation except in a couple [special cases](https://github.com/facebook/infer/blob/85b8ad463cb72fc81edf04261b2aca31b8dfd467/infer/src/clang/cTrans_utils.ml#L556-L559), seemingly because adding them would [break existing checkers](https://github.com/facebook/infer/commit/1486a5f105fe131f93cd3ac48a6f23a0ba6d4ba7). This change fixes many of the casting cases that used to result in information loss without adding `Exp.Cast` expressions. When an `ImplicitCastExpr` has the `part_of_explicit_cast` bit set, it will be forwarded the type of its outer cast. This is particularly useful with `LValueToRValue` casts because it means the dereference assignment generated has the correct type.

Here are a couple cases where this prevents information loss (included as unit tests):

```c
void integral_cast() {
    int a;
    int b = ((char) a) + 2;
}

struct object {
    int field;
};

void pointer_cast() {
    void *obj;
    int f = ((struct object *) obj)->field;
}
```

Here are the clang ASTs for the casts:

```
|-ImplicitCastExpr 0x55e4e4986500 <col:13, col:22> 'int' <IntegralCast>
| `-ParenExpr 0x55e4e49864c0 <col:13, col:22> 'char'
|   `-CStyleCastExpr 0x55e4e4986498 <col:14, col:21> 'char' <IntegralCast>
|     `-ImplicitCastExpr 0x55e4e4986480 <col:21> 'int' <LValueToRValue> part_of_explicit_cast
|       `-DeclRefExpr 0x55e4e4986450 <col:21> 'int' lvalue Var 0x55e4e4986350 'a' 'int'
```

```
`-MemberExpr 0x55e4e49869a8 <col:13, col:38> 'int' lvalue ->field 0x55e4e4986630
  `-ParenExpr 0x55e4e4986988 <col:13, col:35> 'struct object *'
    `-CStyleCastExpr 0x55e4e4986960 <col:14, col:32> 'struct object *' <BitCast>
      `-ImplicitCastExpr 0x55e4e4986948 <col:32> 'void *' <LValueToRValue> part_of_explicit_cast
        `-DeclRefExpr 0x55e4e4986858 <col:32> 'void *' lvalue Var 0x55e4e4986758 'obj' 'void *'
```

Without this change, the `integral_cast` assignment has nothing indicating `a` got truncated to a `char`:

```
n$0=*&a:int
*&b:int=(n$0 + 2)
```

With this change, the assignment has the correct type:

```
n$0=*&a:char
*&b:int=(n$0 + 2)
```

The `pointer_cast` assignment without this change does not give any information about what type the field access is on:

```
n$0=*&obj:void*
n$1=*n$0.field:int
*&f:int=n$1
```

(edit: I realized that there actually is a `typ` stored in `Exp.Lfield`, but it only gives the struct type, not the pointer type, so it's still not ideal. For example, here you'd only have `void*` and `object`, and would have to reconstruct the `object*` yourself. So it's not strictly information loss, but this is much nicer.)

After this change, it has the correct type, allowing checkers to do things with that information:

```
n$0=*&obj:object*
n$1=*n$0.field:int
*&f:int=n$1
```

### DIFF
--- a/facebook-clang-plugins/clang-ocaml/clang_ast_proj.ml.p
+++ b/facebook-clang-plugins/clang-ocaml/clang_ast_proj.ml.p
@@ -218,6 +218,7 @@ let get_cast_kind = function
 #define ABSTRACT_STMT(TYPE)
 #define CASTEXPR(Type, Base) | Type (_, _, _, cast_expr_info)
 #define EXPLICITCASTEXPR(Type, Base) | Type (_, _, _, cast_expr_info, _)
+#define IMPLICITCASTEXPR(Type, Base) | Type (_, _, _, cast_expr_info, _)
 #define CXXNAMEDCASTEXPR(Type, Base) | Type (_, _, _, cast_expr_info, _, _)
 #define OBJCBRIDGEDCASTEXPR(Type, Base) | Type (_, _, _, cast_expr_info, _, _)
 #include<clang/AST/StmtNodes.inc>
--- a/facebook-clang-plugins/libtooling/ASTExporter.h
+++ b/facebook-clang-plugins/libtooling/ASTExporter.h
@@ -390,6 +390,7 @@ class ASTExporter : public ConstDeclVisitor<ASTExporter<ATDWriter>>,
   DECLARE_VISITOR(Expr)
   DECLARE_VISITOR(CastExpr)
   DECLARE_VISITOR(ExplicitCastExpr)
+  DECLARE_VISITOR(ImplicitCastExpr)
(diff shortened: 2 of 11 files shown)


jvillard:
Thank you for your contribution! Sorry it took so long to get merged, just slipped out of my mind over the summer...

PR #1466, opened 2021-07-08
Fix linguist tool configuration (exclude docs from stats)

prekel:
With this change, language stats will be without html:

image


jvillard:
Thank you @prekel, now Infer is in OCaml again :camel:

PR #1482, opened 2021-08-18
add additional models for nullsafe and biabduction

cottamz:

java.util.Objects should be modeled in nullsafe:
https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html


jvillard:
Thank you for your PR, this looks great. Let me know if you have issues with signing the CLA.

I don't think we need the extra files in models/ for your purposes as these are only taken into account by the biabduction analysis. That said, it doesn't hurt to have them around too so up to you whether to leave them in or not.

cottamz:
> Thank you for your PR, this looks great. Let me know if you have issues with signing the CLA.
The CLA is now signed.

> I don't think we need the extra files in models/ for your purposes as these are only taken into account by the biabduction analysis. That said, it doesn't hurt to have them around too so up to you whether to leave them in or not.
I updated the pull request title indicating we are adding models for both nullsafe and biabduction.

cottamz:
Could you share the Phabricator build and test output?

jvillard:
This looks good overall but the changes in the biabduction models had surprising effects on other analyses, we're trying to figure out if we could limit these effects to the biabduction analysis.

davidpichardie:
I will close this one because it is too old now

Issue #1274, opened 2020-06-02
SARIF (Static Analysis Results Interchange Format) format support


XVilka:
It would amazing to have [SARIF](http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html) (Static Analysis Results Interchange Format) format support, so, for example, Infer can be integrated right into the GitHub: https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/uploading-a-sarif-file-to-github

After the upload to GitHub it would look similar to
![image](https://user-images.githubusercontent.com/203261/83507266-e93ba180-a4fa-11ea-8fb9-5febb8b1e6e4.png)

See also their repository: https://github.com/oasis-tcs/sarif-spec

XVilka:
Seems that @GrammaTech implemented SARIF converted for Infer already, but I wasn't able to find any reference:
> As mentioned above, we have written adapters that convert Plist-form (Clang Static Analyzer and Cppcheck), and Facebook Infer output to SARIF. These have been contributed to the SARIF SDK on Github.

https://blogs.grammatech.com/static-analysis-results-a-format-and-a-protocol-sarif-sasp

XVilka:
See also the example how they added SARIF output to the Clang Analyzer natively: https://reviews.llvm.org/D53814

andrewvaughanj:
Looks like the SARIF SDK one is here: https://github.com/sarif-standard/sarif-sdk-typescript/blob/master/InferConverter.ts

xi-liu-ds:
I am contributing a native SARIF exporter in this [PR](https://github.com/facebook/infer/pull/1487)

jvillard:
fixed in #1487

jiangxincode:
Preview unavailable: 
https://github.com/jiangxincode/ApkToolBoxGUI/security/code-scanning

PR #1487, opened 2021-08-27
SARIF export support

xi-liu-ds:
Hi there 👋

As it was pointed out in [this issue](https://github.com/facebook/infer/issues/1274), SARIF format is now widely used in many popular bug filing pipelines. To make it easier to deploy Infer in those popular bug filing pipelines, we decide to create this native support of exporting bug report in SARIF format. 

If you need a bug report in SARIF format, all you need is adding `--output-sarif` flag in your Infer command.

Please review this PR and provide your feedback about this small feature.

Thanks!


infer/src/integration/JsonReports.ml
```diff
@@ -71,6 +73,43 @@ let loc_trace_to_jsonbug_record trace_list ekind =
...
+      in
+      let file_location_to_record trace_item = 
+        {message= message trace_item
+        ; Sarifbug_j.physicalLocation=physical_location trace_item}
+      in
+      let trace_item_to_record trace_item = 
+        { Sarifbug_j.nestingLevel = trace_item.Errlog.lt_level
+        ; location= file_location_to_record trace_item }
+      in
+      let record_list = List.rev (List.rev_map ~f:trace_item_to_record trace_list) in
```
jvillard, line 109: Could be just `List.map`

infer/src/integration/JsonReports.ml
```diff
@@ -140,6 +178,80 @@ end) : Printer with type elt = P.elt = struct
...
+  val pp_results : F.formatter -> unit -> unit
+
+  val pp_close : F.formatter -> unit -> unit
+
+  val pp_rules : F.formatter -> elt -> unit
+
+  val pp : F.formatter -> elt -> unit
+end
+
+module MakeSarifListPrinter (P : sig
```
jvillard, line 195: This functor has only one instantiation, could it be simply a module instead? I.e. inline the functions in `PrinterSarif`.

infer/src/integration/JsonReports.ml
```diff
@@ -140,6 +178,80 @@ end) : Printer with type elt = P.elt = struct
...
+  type elt
+
+  val rules_to_string : elt -> string option
+
+  val to_string : elt -> string option
+end) : PrinterSarif with type elt = P.elt = struct
+  include P
```
jvillard, line 202: I think it's enough to `open P` here, but if this is converted to a module then the functions in the argument of `SarifIssuePrinter` will go here instead of that anyway.

infer/src/integration/JsonReports.ml
```diff
@@ -140,6 +178,80 @@ end) : Printer with type elt = P.elt = struct
...
+    F.fprintf fmt "%t%s: %s,@\n"  (pp_n_spaces 10) "\"informationUri\"" "\"https://github.com/facebook/infer\"" ;
+    F.fprintf fmt "%t%s: \"%d.%d.%d\",@\n"  (pp_n_spaces 10) "\"version\"" Version.major Version.minor Version.patch ;
+    F.fprintf fmt "%t%s: [@?"  (pp_n_spaces 10) "\"rules\""
+
+  let pp_results fmt () =
+    is_first_item := true ;
+    F.fprintf fmt "]@\n" ;
+    F.fprintf fmt "%t%s@\n"  (pp_n_spaces 8) "}" ;
+    F.fprintf fmt "%t%s,@\n"  (pp_n_spaces 6) "}" ;
+    F.fprintf fmt "%t%s: [@?"  (pp_n_spaces 6) "\"results\""
```
jvillard, line 230: Flushing (with `@?`) is expensive and it's enough to do it only once at the end, probably all of the `@?` in this module can be deleted?

infer/src/integration/JsonReports.ml
```diff
@@ -140,6 +178,80 @@ end) : Printer with type elt = P.elt = struct
...
+
+  let pp_close fmt () = 
+    F.fprintf fmt "]@\n" ;
+    F.fprintf fmt "%t%s@\n" (pp_n_spaces 4) "}" ;
+    F.fprintf fmt "%t%s@\n" (pp_n_spaces 2) "]" ;
+    F.fprintf fmt "}@\n@?"
```
jvillard, line 236: Instead of formatting json by hand here, would it make sense to add these types to the .atd file and use the auto-generated code to print these fields?

infer/src/integration/JsonReports.ml
```diff
@@ -241,6 +353,132 @@ module IssuesJson = struct
...
+  let pp_issues_of_error_log fmt error_filter _ proc_loc_opt proc_name err_log =
+    Errlog.iter
+      (fun err_key err_data -> pp fmt {error_filter; proc_name; proc_loc_opt; err_key; err_data})
+      err_log
+  
+  (** Write bug rules in SARIF 2.1 format *)
+  let pp_rules_of_error_log fmt error_filter _ proc_loc_opt proc_name err_log =
+    Errlog.iter
+      (fun err_key err_data -> pp_rules fmt {error_filter; proc_name; proc_loc_opt; err_key; err_data})
+      err_log
```
jvillard, line 479: Instead of starting from `Errlog`, would it be possible to start from report.json? The json report should be the source of truth as it has some additional filtering and tweaking. Starting from the json will avoid having duplicate code between the json output and the sarif output, eg `should_report_proc_name`, the special case for resource leaks, etc. Other reports emitted also start from the json output for the same reasons, see for example `TextReport` and `XMLReport`.

infer/src/integration/JsonReports.ml
```diff
@@ -419,6 +668,35 @@ let process_all_summaries_and_issues ~issues_outf ~costs_outf ~config_impact_out
       IssueLog.load dir_name |> IssueLog.iter ~f:(write_lint_issues filters issues_outf linereader) ) ;
   ()
 
+let process_issues_for_sarif ~issues_outf =
```
jvillard, line 671: This can also be avoided by starting from the json report instead of summaries.

jvillard (commented):
Thanks a lot for this PR, I'm so happy to see SARIF support coming to Infer! I commented inline but my main point would be to start from report.json instead of creating the report "from scratch" by iterating on the summaries. The way it currently works for other report formats is that only report.json does the heavy lifting of going through summaries to collect errors and do a final filtering and massaging of the issues, then all other reports are derived from report.json. It would be nice if it could be the same for the SARIF report. Let me know if you have concerns about that approach.

xi-liu-ds:
> Thanks a lot for this PR, I'm so happy to see SARIF support coming to Infer! I commented inline but my main point would be to start from report.json instead of creating the report "from scratch" by iterating on the summaries. The way it currently works for other report formats is that only report.json does the heavy lifting of going through summaries to collect errors and do a final filtering and massaging of the issues, then all other reports are derived from report.json. It would be nice if it could be the same for the SARIF report. Let me know if you have concerns about that approach.

I think it is a great idea to derive report.sarif from report.json. We are working on implementing this idea and address other comments from you. Thank you so much for your comments and suggestion!

xi-liu-ds:
Hi there 

I have updated this PR according to the suggestion. Now report.sarif is derived from report.json. 

It is ready for another round of review. 

Thanks!

xi-liu-ds:
Hi,

Just want to call out again that this PR is ready for another round of review. Please take a look and let me know if there is any other suggestion.

Thanks!

jvillard:
Thank you for this valuable contribution, @xi-liu-ds!

PR #1538, opened 2021-09-29
[pulse] Fix use of is_functional in unknown_call function

danielmercier:
is_functional in PulseCallOperations.unknown_call is set but never used.


jvillard:
Thank you @danielmercier, great catch!

PR #1545, opened 2021-10-14
.NET resource leak detection exception flow support

xi-liu-ds:
Hi!

This PR adds .NET resource leak exception flow support.

Please let me know what you think.

Thanks!


xi-liu-ds:
Hi,

This PR has been ready for a while. It would be great if anyone can give it another round of review.

Thanks!

infer/src/absint/ProcCfg.ml
```diff
@@ -201,6 +201,98 @@ module Normal = struct
   let wto = Procdesc.get_wto
 end
 
+(** Forward .NET CFG with exceptional control-flow *)
+module DOTNETExceptional = struct
```
jvillard, line 205: It doesn't seem to me that this is specific to dotnet, is there a better name that would capture which exceptional edges are followed? `ExceptionalThrowOnly`? (see below)

infer/src/absint/ProcCfg.ml
```diff
@@ -201,6 +201,98 @@ module Normal = struct
...
+  let fold_normal_or_exn_succs fold_normal_alpha fold_exceptional t n ~init ~f =
+    let choose_normal_or_exn_succs node = 
+      let instrs = Procdesc.Node.get_instrs node in
+      let instr_count = (Instrs.count instrs) - 3 in
+      let get_last_instr () =
+         if instr_count >= 0 then Instrs.nth_exn instrs instr_count else Instrs.last instrs |> Option.value ~default:Sil.skip_instr
+      in
+      let last_instr = get_last_instr () in
+      match last_instr with 
+      | Sil.Store {e1= Exp.Lvar pvar; e2= Exp.Exn _} when Pvar.is_return pvar ->
```
jvillard, line 238: This corresponds to calls to `throw e` that are translated by the (Java and C#?) frontends as just `ret = Exn e`, right? Worth adding a comment about what this is meant to capture and also that this is the only difference with `Exceptional` (unless I missed other changes) as there is an unfortunate amount of code duplication.

jvillard (commented):
Looks good to me, I would just like to double check my understanding first, see comments. Sorry for the delay in reviewing.

We are currently improving the AI framework so that checkers can use it to handle exceptions better. The current `Exceptional` CFG exploration is very naive indeed as it doesn't let checkers distinguish exceptional control flow from normal control flow. Stay tuned! In the meantime this is a good workaround.

infer/src/absint/ProcCfg.ml
xi-liu-ds: Done.

xi-liu-ds: Done

infer/src/absint/ProcCfg.ml
xi-liu-ds: Done. Thanks for suggestion.

infer/src/absint/ProcCfg.ml
xi-liu-ds: Yeah, maybe we can create an Exceptional interface and inherit? In this way we can probably avoid most of the code duplication. Please let me know what you think.

infer/src/absint/ProcCfg.ml
```diff
@@ -201,6 +201,100 @@ module Normal = struct
...
+        List.fold exn_preds ~init ~f
+    | exception Caml.Not_found ->
+        init
+
+  (** we fold the exception flow only when the throw instruction is encountered and exn is returned. 
+  Under this circumstances, resources could be disposed in finaly black through exception sink node *)
+  let fold_normal_or_exn_succs fold_normal_alpha fold_exceptional t n ~init ~f =
+    let choose_normal_or_exn_succs node = 
+      let instrs = Procdesc.Node.get_instrs node in
+      let instr_count = (Instrs.count instrs) - 3 in
```
skcho, line 234: Could we have a comment about `- 3` here?

infer/src/absint/ProcCfg.ml
```diff
@@ -201,6 +201,100 @@ module Normal = struct
   let wto = Procdesc.get_wto
 end
 
+(** Forward CFG with exceptional control-flow for throw exception instruction only*)
+module ExceptionalThrowOnly = struct
+  module Node = DefaultNode
```
skcho, line 206: Can we use `include Exceptional` to avoid the duplicated definitions? We can remove most of them except `fold_normal_or_exn_succs`, `fold_succs`, and `fold_preds` that have different definitions.

infer/src/absint/ProcCfg.ml
xi-liu-ds: Throw exception CFG node actually contains the following three instructions:

image

The third from the last instruction is the one we want to verify.

I have added a comment to explain the intention here.

infer/src/absint/ProcCfg.ml
xi-liu-ds: Done. Thanks for suggestion!

infer/src/absint/ProcCfg.ml
skcho: Thank you for the explanation.

PR #1549, opened 2021-10-26
Enable WSL2 to use daemon

xi-liu-ds:
Hi Infer team,

Since WSL2 now supports daemon now, we remove this restriction to use daemon in WSL2 for faster infer analysis in WSL2 platform. 

Thanks


jvillard:
Thanks a lot @xi-liu-ds!

yakobowski:
Unless someone objects, I will reintroduce the functionality of this commit -- but limit it to WSL1. It seems WSL1 and WSL2 are meant to cohabit, and we just had a customer under WSL1 for which Infer was crashing (because of this PR).

xi-liu-ds:
> Unless someone objects, I will reintroduce the functionality of this commit -- but limit it to WSL1. It seems WSL1 and WSL2 are meant to cohabit, and we just had a customer under WSL1 for which Infer was crashing (because of this PR).

We are good with that if your reintroduction will be limited to WSL1 only. We will need daemon under WSL2 only.

PR #1556, opened 2021-11-12
Detect whether DBWriter can actually be used

yakobowski:
... on systems where it is assumed available.

We have had customers under Linux with DBWriter related errors. Some
filesystems (e.g. ClearCase, or Vagrant synced folders) do not allow
creating a Unix domain socket on them. This patch tries to detect
whether sockets are supported, and falls back to --no-dbwriter
behavior when they are not.

### DIFF
--- a/infer/src/base/DBWriter.ml
+++ b/infer/src/base/DBWriter.ml
@@ -363,8 +363,9 @@ module Server = struct
 
   let socket_exists () = in_results_dir ~f:(fun () -> Sys.file_exists_exn socket_name)
 
-  let server () =
-    L.debug Analysis Quiet "Sqlite write daemon: starting up@." ;
+  (* Error recuperation is done by attempting this function at module initialization time, and
+     not using DbWriter at all in case it fails. See {!can_use_socket} below. *)
+  let setup_socket () =
     if socket_exists () then L.die InternalError "Sqlite write daemon: socket already exists@." ;
     let socket = Unix.socket ~domain:socket_domain ~kind:Unix.SOCK_STREAM ~protocol:0 () in
     in_results_dir ~f:(fun () -> Unix.bind socket ~addr:socket_addr) ;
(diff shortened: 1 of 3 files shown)


infer/src/base/DBWriter.ml
```diff
@@ -363,22 +363,43 @@ module Server = struct
...
+        Unix.close socket ;
+        Unix.unlink socket_name )
+
+
+  (* Check whether we can create a socket to communicate with the asynchronous DBWriter process. *)
+  let can_use_socket () =
+    try
+      (* This function is called very early, and the infer-out directory may not have been created
+         yet. Ensure it exists before attempting to create the socket. *)
+      Unix.mkdir_p Config.toplevel_results_dir ;
```
jvillard, line 390: That's a bit too disgusting :) Could we have `use_daemon` be a lazy value instead?

infer/src/base/DBWriter.ml
```diff
@@ -420,14 +441,27 @@ end
...
+    else Server.socket_exists ()
+  else false
 
 
 let perform cmd = if use_daemon then Server.send cmd else Command.execute cmd
 
 let start () = Server.start ()
 
-let stop () = Server.send Command.Terminate
+let stop () = try Server.send Command.Terminate with Unix.Unix_error _ -> ()
```
jvillard, line 464: Is there a separate issue that the socket creation could succeed but this could fail? As far as I can tell this is only called when `use_daemon` is true.

infer/src/base/DBWriter.ml
yakobowski: Good point, I suspect so. Will experiment and report.

infer/src/base/DBWriter.ml
yakobowski: I'm not 100% sure anymore. To be honest, I wondered about that too when I prepared the code. I will have another look.

yakobowski: Ah actually, I have a nice commit message about exactly this:
```
    Intercept error when shutting down db daemon
    
    Results in much nicer behavior when Infer is interrupted.
```
I will experiment with soft and hard kill and report on the differences with and without this change.

jvillard: If it's a pain to check we can just leave this in.

yakobowski: Without this patch, this is why I currently get when I Ctrl-C a capture that uses DBWriter:
```
465395: Error while running epilogue "Stop Sqlite write daemon":
Unix.Unix_error(Unix.ENOENT, "connect", "").
Powering through...
```
This may be due to our patches for Ada, but I don't think so.
Note that I would be fine with limiting the catch-all to those arguments if you think this helps.

infer/src/base/DBWriter.ml
yakobowski: It works \o/.

yakobowski:
Update following the review pushed.

PR #1564, opened 2021-11-30
SARIF improvements

xi-liu-ds:
Hi all,

This PR improves the usability of SARIF format in deployment by introducing the following new features:

1. Add help URIs to the supported bug types
2. Append "file:" in front of the relative uri of the bug location

### DIFF
--- a/infer/src/integration/SarifReport.ml
+++ b/infer/src/integration/SarifReport.ml
@@ -19,7 +19,18 @@ module ReportSummary = struct
   let pp fmt {n_issues= _; issue_type_counts} =
     let string_of_issue ~issue_type ~issue_type_hum =
       let shortDescription = {Sarifbug_j.text= issue_type_hum} in
-      let rule = {Sarifbug_j.id= issue_type; shortDescription} in
+      let help_uri = 
+        match issue_type with
+        | "DOTNET_RESOURCE_LEAK" ->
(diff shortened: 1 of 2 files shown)


infer/src/integration/SarifReport.ml
```diff
@@ -19,7 +19,18 @@ module ReportSummary = struct
   let pp fmt {n_issues= _; issue_type_counts} =
     let string_of_issue ~issue_type ~issue_type_hum =
       let shortDescription = {Sarifbug_j.text= issue_type_hum} in
-      let rule = {Sarifbug_j.id= issue_type; shortDescription} in
+      let help_uri =
```
jvillard, line 22: Better compute this accurately for all issue types instead. You can get the corresponding IssueType.t from their unique id (which are the strings you have here) using `IssueType.find_from_string`, then generate the URL by exposing `abs_url_of_issue_type` currently in Help.ml so that it's all in one place (that function can stay in Help.ml, you just need to add it to Help.mli):

https://github.com/facebook/infer/blob/9f75f8017f4a9090be4f8057ea06e7118c5ba6ba/infer/src/integration/Help.ml#L24-L25

xi-liu-ds: Done. Thanks for pointing this out!

jvillard:
Thank you!

Issue #1567, opened 2021-12-10
Disable outputing bug traces in infer bug report


xi-liu-ds:
Hi,

I am wondering is it possible (a command flag or something) to disable outputting bug traces, such as those shown in the screenshot below, in infer bug report?

![image](https://user-images.githubusercontent.com/63893533/145629887-01763149-e2d6-45ea-864e-d86d6717363d.png)

Thanks!

jvillard:
There is no such option at the moment. Can you say more about why you'd want to do this? You can post-process the json to get rid of them (for example using [`jq`](https://stedolan.github.io/jq/)) but these bug traces are generally needed to understand infer reports.

xi-liu-ds:
Currently Sarif Viewer extension has some incompatibility with SARIF files with bug traces. We have to disable it to unblock us from deploying InferSharp on some platforms. I have created [this PR](https://github.com/facebook/infer/pull/1569) to disable outputting bug traces in SARIF files through a `-no-sarif-codeflows` flag.

jvillard:
PR has been closed, issue seems resolved.

PR #1569, opened 2021-12-13
Add no code flow flag for SARIF

xi-liu-ds:
Hi,

As we mentioned in [this issue](https://github.com/facebook/infer/issues/1567), we added a new `no-sarif-codeflows` flag to avoid outputting bug traces in SARIF.  

Please ;let us know what you think :)

### DIFF
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -2381,6 +2381,12 @@ and sarif =
     "Output issues in SARIF (Static Analysis Results Interchange Format) in infer-out/report.sarif"
 
 
+and no_sarif_codeflows =
+  CLOpt.mk_bool ~long:"no-sarif-codeflows" ~default:false
+    ~in_help:InferCommand.[(Run, manual_generic)]
+    "Output issues in SARIF (Static Analysis Results Interchange Format) in infer-out/report.sarif without codeflows"
+
+
 and scheduler =
   CLOpt.mk_symbol ~long:"scheduler" ~default:File ~eq:equal_scheduler
     ~in_help:InferCommand.[(Analyze, manual_generic)]
@@ -3557,6 +3563,8 @@ and reports_include_ml_loc = !reports_include_ml_loc
 and results_dir = !results_dir
 
 and sarif = !sarif
+
+and no_sarif_codeflows = !no_sarif_codeflows
 
 and scheduler = !scheduler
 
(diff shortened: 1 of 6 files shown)


infer/src/base/Config.ml
```diff
@@ -2381,6 +2381,12 @@ and sarif =
     "Output issues in SARIF (Static Analysis Results Interchange Format) in infer-out/report.sarif"
 
 
+and no_sarif_codeflows =
+  CLOpt.mk_bool ~long:"no-sarif-codeflows" ~default:false
```
jvillard, line 2385: Can this be called `sarif-codeflows` instead and default to `true`? This would avoid having a `--no-no-sarif-codeflows` option ;)

jvillard:
Looks ok to me in principle. Could you run `make manuals` to update the checked in copies of the manuals (the tests rely on it) as well as `make fmt` to make sure the code is formatted?

xi-liu-ds:
> Looks ok to me in principle. Could you run `make manuals` to update the checked in copies of the manuals (the tests rely on it) as well as `make fmt` to make sure the code is formatted?

I have updated manuals. However, I still have some difficulties in making `make fmt` works (I am developing in a docker container, not sure if that is the reason). Is it easier for you to format at your end? Thanks!

infer/src/base/Config.ml
jvillard: Actually there's logic in CommandLineOption.ml to avoid generating "no-no-xxx" options but still it would be more direct (and more consistent with the other bool options) to define the positive option instead of the negative one.

xi-liu-ds: Sure. Let me update it to `drop-sarif-codeflows`

xi-liu-ds: Done

jvillard: Can this be `--sarif-codeflows` instead? One would pass `--no-sarif-codeflows` to disable it as in your original change, which I thought was good, I just want to avoid a negative definition (to avoid double negations in the code, eg `not Config.no_sarif_codeflows`).

xi-liu-ds:
Hi all!

Thanks for reviewing this PR! Just FYI that this PR may be abandoned since our tooling team is working on a universal way to render code flows. Will keep you posted when the final decision is made.

xi-liu-ds:
> Hi all!
> 
> Thanks for reviewing this PR! Just FYI that this PR may be abandoned since our tooling team is working on a universal way to render code flows. Will keep you posted when the final decision is made.

It is done. Let me close this PR. Let me know if this is still useful and I can reopen it. Thanks!

PR #1594, opened 2022-01-26
Fix cache name in CI

yakobowski:
runner.os is actually a very uninformative name: either 'Linux' or 'macOS'.
When ubuntu-latest or macOS-latest changes, the cached binaries become
incompatible with the new OS. But this is not detected by the cache
key we use. Thus, use 'os-version-action' to get this information.

Also, use 'matrix.os' instead of 'runner.os', which is slightly more
precise.  This changes nothing in practice, since Infer is tested on
only one version of each OS for the moment.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -73,6 +73,10 @@ jobs:
         run: brew install automake jq ninja pkg-config lzlib zlib coreutils
         if: runner.os == 'macOS'
 
+      - name: Get OS version
+        uses: sersoft-gmbh/os-version-action@v1
+        id: os-version
+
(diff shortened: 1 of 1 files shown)


jvillard:
Thanks! Is this still work in progress?

yakobowski:
Yes and no :grimacing:.  It fixes a real bug, in the sense that the action will bug when ubuntu-latest will be updated. OTOH it does not fix the MacOS bug, and I don't understand why.

I won't have time to look at this anytime soon, especially since we don't use MacOS here. So this could as well be merged.

jvillard:
I guess this still makes things strictly better than before so in it goes!

PR #1597, opened 2022-02-01
Incremental build support for json-based frontend

xi-liu-ds:
Hi!

In this PR we add incremental build support for json-based frontend. All we need is provide a list of changed files in a txt file and feed in through flag `--changed-files-index` for `infer analyzejson` command.

Please let us know your thoughts. Thanks!


jvillard:
Hi @xi-liu-ds, sorry for the delay. It looks like this duplicates more code still from InferAnalyze.ml. Would that duplication still be needed if parsing the json was done in a capture phase instead of together with the analysis? It feels to me like doing so would give you incremental analysis for free as well as being more maintainable, as the analysis phase would be just the normal Infer analysis. Also there wouldn't need to be an `analyzejson` command.

I think capturing json could be done similarly as how we capture clang compilation databases so you could take inspiration from those code paths. We use it like this:

```
# does capture + analysis
$ infer run --compilation-database file.json
# equivalent to running these two commands:
$ infer capture --compilation-database file.json
$ infer analyze
```

Similarly we'd have two options `--json-cfg` and `--json-tenv` that have to be specified together or not at all:
```
# does capture + analysis
$ infer run --json-cfg cfg.json --json-tenv tenv.json
# equivalent to running these two commands:
$ infer capture --json-cfg cfg.json --json-tenv tenv.json
$ infer analyze
```

So we would have a new [driver mode](https://github.com/facebook/infer/blob/cfdcf8322c6c2b73609c0f6675c2eaaf1502948e/infer/src/integration/Driver.ml#L15) `JsonSIL` set similarly to [ClangCompilationDB](https://github.com/facebook/infer/blob/cfdcf8322c6c2b73609c0f6675c2eaaf1502948e/infer/src/integration/Driver.ml#L374-L377). Then it's just a matter of separating out the capture phase out of InferAnalyzeJson.ml and into its own module and call it from Driver.ml, then the analysis phase can be deleted from InferAnalyzeJson.ml (and so the rest of the file with it) as it's not a separate thing from the normal infer analysis anymore.

Would that approach work for you and are there any blockers? I'm not sure how much we need to ensure backwards compability or we can just get rid of the `analyzejson` command altogether if we go down this route, for example.

xi-liu-ds:
> Hi @xi-liu-ds, sorry for the delay. It looks like this duplicates more code still from InferAnalyze.ml. Would that duplication still be needed if parsing the json was done in a capture phase instead of together with the analysis? It feels to me like doing so would give you incremental analysis for free as well as being more maintainable, as the analysis phase would be just the normal Infer analysis. Also there wouldn't need to be an `analyzejson` command.
> 
> I think capturing json could be done similarly as how we capture clang compilation databases so you could take inspiration from those code paths. We use it like this:
> 
> ```
> # does capture + analysis
> $ infer run --compilation-database file.json
> # equivalent to running these two commands:
> $ infer capture --compilation-database file.json
> $ infer analyze
> ```
> 
> Similarly we'd have two options `--json-cfg` and `--json-tenv` that have to be specified together or not at all:
> 
> ```
> # does capture + analysis
> $ infer run --json-cfg cfg.json --json-tenv tenv.json
> # equivalent to running these two commands:
> $ infer capture --json-cfg cfg.json --json-tenv tenv.json
> $ infer analyze
> ```
> 
> So we would have a new [driver mode](https://github.com/facebook/infer/blob/cfdcf8322c6c2b73609c0f6675c2eaaf1502948e/infer/src/integration/Driver.ml#L15) `JsonSIL` set similarly to [ClangCompilationDB](https://github.com/facebook/infer/blob/cfdcf8322c6c2b73609c0f6675c2eaaf1502948e/infer/src/integration/Driver.ml#L374-L377). Then it's just a matter of separating out the capture phase out of InferAnalyzeJson.ml and into its own module and call it from Driver.ml, then the analysis phase can be deleted from InferAnalyzeJson.ml (and so the rest of the file with it) as it's not a separate thing from the normal infer analysis anymore.
> 
> Would that approach work for you and are there any blockers? I'm not sure how much we need to ensure backwards compatibility or we can just get rid of the `analyzejson` command altogether if we go down this route, for example.

Thanks Jules for the suggestion! We are aware that analyze json has some code duplication from `InferAnalyze.ml`. We are investigating the solution you provided and see if it can fully replace the current experience from `InferAnalyzeJson.ml`.

xi-liu-ds:
Hi @jvillard So sorry about the delay on this PR. I finally had chance to look deep into it this week.

It turns out that your suggestions are great and helped to remove lots of duplicate code! Thank you so much! 

Like you suggested, now we can do `infer capture --json-cfg cfg.json --json-tenv tenv.json` to run json frontend analysis. And for incremental build, all we need is `infer capture --json-cfg cfg.json --json-tenv tenv.json --incremental-analysis --changed-files-index change_file_index.txt`

Please take a look when you get a chance. Thanks again!

infer/src/integration/Driver.ml
```diff
@@ -60,6 +58,8 @@ let pp_mode fmt = function
       F.fprintf fmt "Gradle driver mode:@\nprog = '%s'@\nargs = %a" prog Pp.cli_args args
   | Javac {prog; args} ->
       F.fprintf fmt "Javac driver mode:@\nprog = '%s'@\nargs = %a" prog Pp.cli_args args
+  | JsonSIL {json_cfg; json_tenv} ->
+      F.fprintf fmt "Analyze json mode"
```
jvillard, line 62: typo: this isn't "analyze" anymore. Also it could be useful to dump the paths here to ease debugging, like the other modes.

infer/src/integration/Driver.ml
```diff
@@ -283,18 +275,16 @@ let analyze_and_report ?suppress_console_report ~changed_files mode =
...
     ResultsDir.RunState.set_merge_capture false ) ;
   if should_analyze then
-    if analyze_json then execute_analyze_json ()
-    else if SourceFiles.is_empty () && Config.capture then error_nothing_to_analyze mode
+    if SourceFiles.is_empty () && Config.capture then error_nothing_to_analyze mode
     else (
       execute_analyze ~changed_files ;
       if Config.starvation_whole_program then StarvationGlobalAnalysis.whole_program_analysis () ) ;
-  if should_report && Config.report then report ?suppress_console:suppress_console_report ()
+  if (should_report || Language.curr_language_is Language.CIL) && Config.report then report ?suppress_console:suppress_console_report ()
```
jvillard, line 287: Why is this needed?

infer/src/integration/Driver.ml
```diff
@@ -390,7 +380,12 @@ let mode_of_build_command build_cmd (buck_mode : BuckMode.t option) =
       if not (List.is_empty Config.clang_compilation_dbs) then (
         assert_supported_mode `Clang "clang compilation database" ;
         ClangCompilationDB {db_files= Config.clang_compilation_dbs} )
-      else Analyze
+      else (
+        match (Config.cfg_json, Config.tenv_json) with
+        | Some cfg_json, Some tenv_json ->
+            JsonSIL {json_cfg= cfg_json; json_tenv= tenv_json}
```
jvillard, line 386: if you bind the values to have the same names as the record fields then you can pun them
```
        | Some json_cfg, Some json_tenv ->
            JsonSIL {json_cfg; json_tenv}
```

infer/src/integration/Driver.ml
```diff
@@ -390,7 +380,12 @@ let mode_of_build_command build_cmd (buck_mode : BuckMode.t option) =
...
+        | _, _ ->
```
jvillard, line 387: you can also write just `| _ ->` for that case

infer/src/integration/CaptureSILJson.ml
```diff
@@ -0,0 +1,28 @@
...
+  let cfg = InferAnalyzeJson.parse_cfg (Yojson.Safe.from_file json_cfg) in
+  let source_file = SourceFile.create ~warn_on_error:false "./Program.cs" in
+  Tenv.store_global tenv ;
+  Language.curr_language := Language.CIL ;
+  SourceFiles.add source_file cfg Tenv.Global None ;
+  (* L.progress "%a@." Cfg.pp_proc_signatures cfg ; *)
+  let exe_env = Exe_env.mk () in
+  Ondemand.analyze_file exe_env source_file ;
+  if Config.write_html then Printer.write_all_html_files source_file ;
+  ()
```
jvillard, line 28: I'm a bit confused as to why you need that part at all (except `Tenv.store_global tenv`). This seems to analyse a fake file "./Program.cs", but why?

infer/src/IR/Attributes.ml
```diff
@@ -94,6 +94,55 @@ let load, clear_cache, store =
...
+    let proc_name = Procname.SQLite.serialize pname in
+    let attr_kind = int64_of_attributes_kind akind in
+    let cfg = Procdesc.SQLite.serialize proc_desc in
+    let proc_attributes = ProcAttributes.SQLite.serialize attr in
+    let source_file = SourceFile.SQLite.serialize attr.loc.Location.file in
+    let callees =
+      Option.value_map proc_desc ~f:Procdesc.get_static_callees ~default:[]
+      |> Procname.SQLiteList.serialize
+    in
+    DBWriter.replace_attributes ~proc_uid ~proc_name ~attr_kind ~source_file ~proc_attributes ~cfg ~callees ;
```
jvillard, line 143: This bypasses and invalidates the caching done above. That logic should be moved in the main `store` function above instead of in a copy of `store` without the cache.

Can you say more about why the changes in this file are needed? Why do you need to update things in the DB after the fact?

jvillard (commented):
Thanks, this looks so much better! I have a few questions.

infer/src/integration/Driver.ml
xi-liu-ds: Good catch! Fixed

infer/src/integration/Driver.ml
xi-liu-ds: Done

infer/src/integration/Driver.ml
xi-liu-ds: Done

infer/src/integration/CaptureSILJson.ml
xi-liu-ds: Because we want to create `SourceFiles` object using cfg. I didn't find a better solution. Could you share if there is a better option?

infer/src/integration/Driver.ml
xi-liu-ds: Removed.

xi-liu-ds:
Thanks for your comments! Now the incremental build support has been improved even more. It can be trigger by using the following commands:

```bash
infer capture --cfg-json cfg.json --tenv-json tenv.json
infer analysis --changed-files-index change_file_index.txt -j 1
```
Or just simply
```bash
infer run --cfg-json cfg.json --tenv-json tenv.json --changed-files-index change_file_index.txt -j 1
```

infer/src/IR/Attributes.ml
xi-liu-ds: Sorry it seems that I have created a duplicate `store` function. Now it has been removed and fixed.

We need to update DB when new procedure names were added in changed files. So Infer can be notified to analyze these added procedures as well

infer/src/integration/CaptureSILJson.ml
xi-liu-ds: A second thought, I think there is [a better solution](https://github.com/facebook/infer/blob/908e770d7323a1cd0b63d74f41c99f0a5ee911f1/infer/src/integration/CaptureSILJson.ml#L19) instead of using a fake file. And it solves issues in incremental mode, too!

infer/src/backend/InferAnalyze.ml
```diff
@@ -261,6 +261,7 @@ let main ~changed_files =
   let source_files = lazy (get_source_files_to_analyze ~changed_files) in
   (* empty all caches to minimize the process heap to have less work to do when forking *)
   clear_caches () ;
+  if Config.cil_lang then Language.curr_language := Language.CIL ;
```
jvillard, line 264: Is this needed? I would think that this is set at the beginning of the analysis of each procedure anyway so we'll get the correct language by the time we start analysing any code. Then we wouldn't need the new config option `Config.cil_lang`.

infer/src/backend/InferAnalyzeJson.ml
```diff
@@ -679,6 +679,28 @@ let parse_cfg (json : Safe.t) =
...
+
+let store cfg =
+  let cfg_map = ref SourceFile.Map.empty in
+  let save_proc procname proc_desc =
+    let attributes = Procdesc.get_attributes proc_desc in
+    let source_file = attributes.loc.Location.file in
+    cfg_map := add_proc_to_cfg source_file procname proc_desc cfg_map
+  in
+  Procname.Hash.iter save_proc cfg ;
+  SourceFile.Map.iter (fun sf mapped_cfg -> SourceFiles.add sf mapped_cfg Tenv.Global None) !cfg_map
```
jvillard, line 701: Seems like this could easily be rewritten without a ref if you would like to simplify this a little bit.

```
let add_proc_to_cfg source_file pname pdesc map =
  match SourceFile.Map.find_opt source_file map with
  | Some cfg ->
      Procname.Hash.add cfg pname pdesc ;
      SourceFile.Map.add source_file cfg map
  | None ->
      let new_cfg = Cfg.create () in
      Procname.Hash.add new_cfg pname pdesc ;
      SourceFile.Map.add source_file new_cfg map


let store cfg =
  let cfg_map = refn
  let save_proc procname proc_desc cfg_map =
    let attributes = Procdesc.get_attributes proc_desc in
    let source_file = attributes.loc.Location.file in
    add_proc_to_cfg source_file procname proc_desc cfg_map
  in
  let cfg_map = Procname.Hash.fold save_proc cfg SourceFile.Map.empty in
  SourceFile.Map.iter (fun sf mapped_cfg -> SourceFiles.add sf mapped_cfg Tenv.Global None) cfg_map
```

infer/src/integration/CaptureSILJson.ml
```diff
@@ -0,0 +1,17 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *)
+
+(** Capture module for the json analysis in the capture phase *)
+
+open! IStd
+
+let capture ~changed_files ~cfg_json ~tenv_json =
+  let tenv = InferAnalyzeJson.parse_tenv (Yojson.Safe.from_file tenv_json) in
```
jvillard, line 13: Maybe this is a good point to set `Language.curr_language` to `CIL` instead of during the analysis, if it's needed at all (does anything break without?). The clang and Java frontends do the same, though I confess I'm not sure why exactly they need to (eg the Erlang frontend doesn't set the current language).

jvillard (commented):
Very nice, just a few minor comments left on my end.

infer/src/integration/CaptureSILJson.ml
xi-liu-ds: Yes, if we don't set `Language.curr_language` to `CIL`, some of our CIL-specific logics will be skipped and ended up with some bug reports missing. 

I found that if we run `infer run` command, set `Language.curr_language` to `CIL` in `CaptureSILJson` will work just fine for us. However, when we split `infer run` into `infer capture` and `infer analyze`, set `Language.curr_language` to `CIL` in  `CaptureSILJson` won't be fine for us because the language set in `infer capture` won't be passed to `infer analyze`. That is the reason why we ended up creating a `--cli-lang` flag. Do you have any better suggestion for this?

infer/src/backend/InferAnalyze.ml
xi-liu-ds: Like I mentioned in the response to another comment, when we split `infer run` into `infer capture` and `infer analyze`, set `Language.curr_language` to `CIL` in  `CaptureSILJson` won't be fine for us because the language set in `infer capture` won't be passed to `infer analyze`. That is the reason why we ended up creating a `--cli-lang` flag. Do you have any better suggestion for this?

infer/src/integration/CaptureSILJson.ml
jvillard: I've done some digging: the issue seems to be in the way that we disable filtering for C# in JsonReports.ml:
https://github.com/facebook/infer/blob/b67e2d518dd67576e248785e72cb9ebd1a508607/infer/src/integration/JsonReports.ml#L75

This relies on the global `curr_language` being set to `CIL` but we are outside any particular analysis here (this is run after all analyses have run and collects all the summaries of all the procedures; conceivably there could be several languages represented here so `Language.curr_language` doesn't make sense any more). We should test the language of the current proc name for which we want to report the issue instead:

```diff
diff --git a/infer/src/integration/JsonReports.ml b/infer/src/integration/JsonReports.ml
index 196c857ea3..9691b211d6 100644
--- a/infer/src/integration/JsonReports.ml
+++ b/infer/src/integration/JsonReports.ml
@@ -71,8 +71,8 @@ let loc_trace_to_jsonbug_record trace_list ekind =
       record_list
 
 
-let should_report issue_type error_desc =
-  if (not Config.filtering) || Language.curr_language_is CIL then true
+let should_report proc_name issue_type error_desc =
+  if (not Config.filtering) || Language.equal (Procname.get_language proc_name) CIL then true
   else
     let issue_type_is_null_deref =
       let null_deref_issue_types =
@@ -187,7 +187,7 @@ module JsonIssuePrinter = MakeJsonListPrinter (struct
     if
       error_filter source_file err_key.issue_type
       && should_report_proc_name
-      && should_report err_key.issue_type err_key.err_desc
+      && should_report proc_name err_key.issue_type err_key.err_desc
       && not (is_in_clang_header source_file)
     then
       let severity = IssueType.string_of_severity err_key.severity in
```

This filtering is only about one special case: null deref issues reported by biabduction, and CIL issues get filtered out because they have no "bucket" set for them. Buckets are a concept in the biabduction analysis used to represent the "confidence" we have in the issue to report, and if no bucket is set then by default we don't report the issue:
https://github.com/facebook/infer/blob/main/infer/src/integration/JsonReports.ml#L84
https://github.com/facebook/infer/blob/main/infer/src/absint/Localise.ml#L86-L89

Could you try the fix above and remove the config option? The tests pass for me if I do this. I'd rather not have the option if we don't need it, and also because infer should work even if we mix several languages in the same analysis and this breaks that (not a deal breaker since I doubt this is very useful at the moment but still, no reason to break it unless we need to).

infer/src/backend/InferAnalyzeJson.ml
xi-liu-ds: Thanks for your suggestion! Done.

infer/src/integration/CaptureSILJson.ml
xi-liu-ds: Thanks for your comments and advice! I have adopted your changes and removed config, which passed our tests as well.

xi-liu-ds:
Fix unit tests since now we merged `infer analyzejson` into `infer run` now.

jvillard:
thanks @xi-liu-ds, very happy to have finally merged this! :tada: 

I also merged InferAnalyzeJson into CaptureSILJson in e6809d9b5196311d6cc68eb409313645da63ef81

PR #1600, opened 2022-02-16
Adding eli5 video

dmitryvinn:
Here at Meta Open Source, we've been creating short-form videos to explain Meta OSS projects in the simplest terms possible. We started embedding these videos into project pages as you can see in [Hermes home page](https://hermesengine.dev/) or in Docusaurus home page](https://docusaurus.io/)

## Test plan
This is how the video appears on the home page:
image

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -66,6 +66,29 @@ const poweredByItems = [
   { url: "https://wolfssl.com", name: "wolfSSL" },
 ];
 
+function VideoContainer() {
+  return (
+    <div className="container text--center margin-bottom--xl margin-top--lg">
+      <div className="row">
+        <div className="col">
+          <h2>Check it out in the intro video</h2>
+          <div className={styles.ytVideo}>
+            <iframe
+              width="560"
+              height="315"
+              src="https://www.youtube.com/embed/swrmPTJAGqQ"
+              title="Explain Like I'm 5: Infer"
+              frameBorder="0"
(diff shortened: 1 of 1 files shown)


website/src/pages/index.js
```diff
@@ -124,6 +147,7 @@ function Home() {
         </div>
       </header>
       <main>
+      <VideoContainer />
```
jvillard, line 150: Can the video be embedded after the features, so around line 164 below? I like how it looks on the Hermes page a little bit better.

jvillard (commented):
Great idea, thank you for doing this, it would have taken me ages to figure out how to do that!

website/src/pages/index.js
dmitryvinn: I am actually changing how it's going to be on Hermes page also to be above Features - it's better when into video comes before those, as many companies like Supabase do it that way also. We are changing to be like in https://docusaurus.io/

jvillard: The Supabase example is interesting because I feel that it's actually closer to how the infer landing page looks like now, with the "demo" video near the top ("Infer in action" subsection). The ELI5 video is awesome but, in my humble opinion, it's more of a "learn more" video than an intro video compared to the "infer in action" one. Curious to hear your thoughts on that. Anyhow, happy to land this as-is now and iterate further after that.

By the way, in the ELI5 video the "Infer in action" widget doesn't render, was there an issue with it? It should be showing https://asciinema.org/a/32101

PR #1601, opened 2022-02-16
Fix dotnet resource leak assign 

xi-liu-ds:
Hi,

In this PR, we fixed an issue in assign instruction, in which the alias of resources released expression was not traced to the original expression with resources acquired. Meanwhile, we add a couple of unit tests to validate this change.

Please take a look and let me know if you have any suggestions or questions.

Thanks!


infer/src/dotnet/ResourceLeakCSDomain.ml
```diff
@@ -118,18 +118,27 @@ let assign lhs_access_path rhs_access_path held =
...
     match
       AccessPath.replace_prefix ~prefix:rhs_access_path ~replace_with:access_path lhs_access_path
     with
-    | Some base_access_path ->
+    | Some base_access_path -> 
         add_type_map access_path base_access_path ;
-        ResourcesHeld.add base_access_path count held
-    | None ->
+        ResourcesHeld.add base_access_path count held 
+    | None ->
```
jvillard, line 133: Minor: These spurious whitespace changes make this harder to review, especially since GitHub does a poor job of displaying them. Can you try installing ocamlformat with `make devsetup` and apply it to future PRs with `make fmt`?

infer/src/dotnet/ResourceLeakCSDomain.ml
```diff
@@ -118,18 +118,27 @@ let assign lhs_access_path rhs_access_path held =
     | exception Caml.Not_found ->
         ()
   in
+  let compare_expr_base ((base1, _)) ((base2, _)) = 
+    if AccessPath.equal_base base1 base2 then true
+    else false
+  in
```
jvillard, line 124: minor: `compare_*` functions in OCaml typically return -1/0/1, this function should be called `equal_expr_base` or just `equal_base`. It can also be simplified:

```
let equal_base (base1, _) (base2, _) = AccessPath.equal_base base1 base2 in
```

infer/src/dotnet/ResourceLeakCSDomain.ml
```diff
@@ -118,18 +118,27 @@ let assign lhs_access_path rhs_access_path held =
...
-        ResourcesHeld.add base_access_path count held
-    | None ->
+        ResourcesHeld.add base_access_path count held 
+    | None -> 
         if AccessPath.equal rhs_access_path access_path then (
           add_type_map access_path lhs_access_path ;
           ResourcesHeld.add lhs_access_path count held )
-        else ResourcesHeld.add access_path count held
+        else if compare_expr_base rhs_access_path access_path then (
+          let new_access_path = construct_access_path lhs_access_path access_path in
```
jvillard, line 138: I could be totally off but shouldn't this do something like `AccessPath.replace_prefix ~prefix:rhs_access_path ~replace_with:lhs_access_path access_path` instead? Eg if the assignment is `x.f = y.g.h` and `access_path` the access path held was `y.g.h.j` then we'd want the result to be `x.f.j` not `x.g.h.j`?

jvillard:
Another drive-by comment as I read this checker a bit more closely: doesn't using a hashtable to store the type map (and others) cause problems with the interprocedural analysis? Usually mutable state needs to be saved and restored whenever a callee gets analysed, see [ondemand.ml](https://github.com/facebook/infer/blob/b06fd22a780b9d56dfd64639804c3b68a14e2e4c/infer/src/backend/ondemand.ml#L94-L133) otherwise analysing callees will pollute the state in the caller. Would it be better to make these tables part of the domain as regular (immutable) maps instead?

infer/src/dotnet/ResourceLeakCSDomain.ml
xi-liu-ds: I have tried as you suggested. However, I got error and `make fmt` failed.
image
Do you know how to fix this?

jvillard: Right, that's this issue, sorry about that:
https://github.com/facebook/infer/blob/84f6b5f25acb574676f81616113b611587580ee0/infer/src/.ocamlformat#L8-L26

Alternatively you can `opam pin ocamlformat.0.19.0-24-g2596a5c7 -k git https://github.com/danielmercier/ocamlformat.git#master --no-action && opam install ocamlformat odoc-parser.0.9.0` but subsequent `make devsetup` will undo it. A new release of dune is coming that will finally get rid of this issue.

infer/src/dotnet/ResourceLeakCSDomain.ml
xi-liu-ds: Good catch! Fixed

infer/src/dotnet/ResourceLeakCSDomain.ml
xi-liu-ds: Done

infer/src/dotnet/ResourceLeakCSDomain.ml
xi-liu-ds: Fixed. Thanks!

jvillard:
Oops, this was merged by mistake in an old version of the PR, sorry.

PR #1609, opened 2022-03-08
Follow up PR of fixing dotnet resource leak assign

xi-liu-ds:
Hi,

We create this PR as a follow up PR to address the comments from a [previous PR](https://github.com/facebook/infer/pull/1601)


xi-liu-ds:
[This comment](https://github.com/facebook/infer/pull/1601#issuecomment-1041453418) is addressed in the new push. Now it is ready for another review.

Thanks!

infer/src/dotnet/ResourceLeakCSDomain.ml
```diff
@@ -73,53 +74,64 @@ let check_count access_path held =
...
+let reset_type_map = type_map := ResourcesTypeHeld.empty
+
+(** Return leaked types as string*)
+let type_to_list held = 
+  let leaked_type_list = ref [] in
+  let concat_types = 
+    ResourcesTypeHeld.iter
+      (fun x y ->
+        if check_count x held then
+          leaked_type_list := LeakList.append_one !leaked_type_list y )
```
jvillard, line 86: Should we worry about the quadratic complexity here? `append_one` appends at the end of `leaked_type_list`, which is linear, and we do that once for each binding in `type_map`, not sure how big that is. Usually it's better to build the list in reverse and reverse it at the end (assuming you care about the order) to do the whole thing in linear time.

infer/src/dotnet/ResourceLeakCSDomain.ml
```diff
@@ -46,10 +45,12 @@ module BoundsWithTop = struct
 end
 
 module ResourcesHeld = AbstractDomain.Map (AccessPath) (BoundsWithTop)
+module ResourcesTypeHeld = PrettyPrintable.MakePPMap (AccessPath)
+
 open AbstractDomain.Types
 
 (** Initializes resources to type map *)
-let type_map = ref (Hashtbl.create 100)
+let type_map = ref ResourcesTypeHeld.empty
```
jvillard, line 53: This is still mutable state, shouldn't it be part of the domain instead? Is it a global ref because it's global to the function or to the checker? If the former, either it should be made part of the domain or it should be saved and reset by [ondemand.ml with the rest of the "global state"](https://github.com/facebook/infer/blob/49ffdcea8b3a178e7fe764e9acbf4a3b0bf8242d/infer/src/backend/ondemand.ml#L79-L92) every time we change function (in particular, when analysing `foo` that calls `bar`, the type map of `foo` will leak into `bar` and be modified in the analysis of `bar` before coming back to `foo`, or is there a reason why this doesn't happen?).

xi-liu-ds: Thanks! However, I am not 100% sure what you are suggesting. I guess we can create the type mapping as an immutable state in two options: 

1. Revise 'ResourceHeld' as mapping ``AccessPath` to two values: `BoundsWithTop` and `Type`. However, I am not sure if it is feasible in ocaml?  
2. Create a separate `ResourceTypeHeld` and initialize it as a passing parameter to `Analyzer.compute_post` (https://github.com/facebook/infer/blob/39604ffab1cb76f3ef54afadebb6030a09f93b41/infer/src/dotnet/ResourceLeaksCS.ml#L184). However, this function does not take a second state. So we may need to overload this function.

Please let me know if I am on the right track.

jvillard: Either of these should be ok. You'll need to turn the type of the values of the current hash table into an abstract domain too, with `AbstractDomain.Flat` or something more fancy (*).
1. There's `AbstractDomain.Pair` to combine `BoundsWithTop` and `Type` together if a given access path always have a type whenever it is mapped to a bound and vice versa: `AbstractDomain.Map (AccessPath) (AbstractDomain.Pair (BoundsWithTop) (AbstractDomain.Flat (Typ.Name)))`
2. Or, if the maps do not necessarily have the same keys bound at all time or the bindings don't change in sync, your entire domain can be a pair eg `AbstractDomain.Pair (ResourcesHeld) (AbstactDomain.Flat (Typ.Name))`


(*) it seems that currently that type is strings, obtained from, eg, `CSharp.get_class_name`. These are typically expensive to build if that happens often, it's usually better to try and keep the original type (eg `Typ.name`) for as long as possible to avoid building those strings unless needed, hence why I  used `Typ.Name` above.

xi-liu-ds: That is very helpful! I have updated according to your suggestions. Please take a look!

infer/src/dotnet/ResourceLeakCSDomain.ml
xi-liu-ds: Done

xi-liu-ds:
In addition, I have improved the readability of the resource leak message. Now it looks like something attached below:

image

xi-liu-ds:
A `make fmt` issue: when I run `make fmt` it reports the following error:

image

FYI, I am developing in a docker container.

xi-liu-ds:
> > A `make fmt` issue: when I run `make fmt` it reports the following error:
> > image
> > FYI, I am developing in a docker container.
> 
> It says that parallel is not found so just install it? :)
> 
> The PR fails the dotnet resource leaks test that's checked in, is it expected? You can test with `make direct_dotnet_resourceleak_test`, it introduces some false positives:
> 
> ```
> --- a/mnt/btrfs/trunk-git-infer-15-1649327228/infer/tests/codetoanalyze/dotnet/resourceleak/issues.exp
> +++ b/mnt/btrfs/trunk-git-infer-15-1649327228/infer/tests/codetoanalyze/dotnet/resourceleak/issues.exp.test
> @@ -1,2 +1,4 @@
> {+codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakInitInterprocOk(), 0, DOTNET_RESOURCE_LEAK, no_bucket+}
> codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakInitInterprocBad(), 0, DOTNET_RESOURCE_LEAK, no_bucket
> {+codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakInterprocOk(), 0, DOTNET_RESOURCE_LEAK, no_bucket+}
> codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakInterprocBad(), 0, DOTNET_RESOURCE_LEAK, no_bucket
> @@ -6,2 +8,3 @@ codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.Res
> codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakStaticSingletonBad(), 0, DOTNET_RESOURCE_LEAK, no_bucket
> {+codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakGlobalResourceInterprocOk(), 0, DOTNET_RESOURCE_LEAK, no_bucket+}
> codetoanalyze/dotnet/resourceleak/ResourceLeak.cs, Void TestCodeResourceLeak.ResourceLeakGlobalResourceInterprocBad(), 0, DOTNET_RESOURCE_LEAK, no_bucket
> ```

Yes, I made a small mistake. Now all the unit tests passed. Meanwhile I have installed parallel and formatted the code. Thanks for finding them out!

PR #1624, opened 2022-04-19
Add ident description

xi-liu-ds:
Hi,

In this PR we add a new attribute called "description" to identifier for better user experience. For example, we will return something like following in resource leak message.

image

The highlighted message is the description of the resource leaked identifier.

We think the whole Infer community could benefit from it. Please let us know your feedbacks! We can do proper changes accordingly. 

Thanks!


jvillard:
I'm a bit worried about the performance implications of such a change. Moreover, except in C#, every ident created will get a "None" description added to it. I think this is more the job of individual analyses to keep track of where values came from to report sensible messages. For example, in Pulse we solve this issue by introducing a "decompiler" that keeps track of where values came from. I would prefer a similar solution here too: you could add an element to your abstract domain that records this kind of information for each ident.

matjin:
We can remove this PR; we switched over to Pulse resource leak.

geralt-encore:
Closing as per @matjin comment

PR #1641, opened 2022-05-20
Handle static closure calls

MaelCravero:
This PR aims at handling static closure calls like regular static function calls. For instance, this allows for static closure calls to be present in the syntactic callgraph (which would not be the case otherwise).


jvillard:
Are the biabduction changes relevant to you? If not it would be simpler to drop them to avoid breaking something there by accident. We don't have large-scale tests for biabduction anymore as we have stopped using it in production (replaced by Pulse).

yakobowski:
The changes in `/biabduction` are only in `Errdesc`, to capture cases where a closure was used instead of a regular function. So it should be very safe. But we have recently abandoned BO on our side, so we could actually drop this patch too.

PR #1667, opened 2022-08-17
Expanding pulse's analysis for CSharp.

archerD:
Primarily adds resource leak detection to pulse's models for CSharp.
Also adds models for collections in CSharp and changes the behavior for unknown functions in CSharp.


infer/src/dotnet/ResourceLeaksCS.ml
```diff
@@ -62,7 +62,7 @@ module TransferFunctions (CFG : ProcCfg.S) = struct
     in
     match instr with
     | Call (_return, Direct callee_procname, HilExp.AccessExpression allocated :: _, _, _loc)
-      when acquires_resource tenv callee_procname && is_not_enumerable ->
+      when acquires_resource tenv callee_procname && is_not_enumerable->
```
jvillard, line 65: Please try to run `make fmt` before submitting PRs to make reviewing easier and so that the CI doesn't fail because of ill-formatted code.

infer/src/pulse/PulseDiagnostic.ml
```diff
@@ -398,6 +404,23 @@ let get_message diagnostic =
       in
       F.asprintf "Resource dynamically allocated %a is not closed after the last access at %a"
         pp_allocation_trace allocation_trace Location.pp location
+  | CSharpResourceLeak {class_name; location; allocation_trace} ->
+      (* NOTE: this is very similar to the MemoryLeak case *)
```
jvillard, line 408: It's even more similar to the JavaResourceLeak case ;) More seriously, it seems that the way CSharp resources are handled is virtually the same as Java (we'd need to make `class_name` be a more general type to accept either Java or CSharp). Is there some more fundamental difference I missed? If not I'd be interested in merging the two together if you provide some test cases for C# to make sure I don't break anything.

matjin: Agree that outside of the models/class_name construct there shouldn't be any difference between CSharp and Java resources. Here is a PR draft documenting test cases+expected results (we didn't get a chance to check them into the InferSharp test suite, but at least had this): https://github.com/microsoft/infersharp/pull/156/files 

I'll separately email you the cfg.json/tenv.json to save you having to build the project/translate!

matjin: This is in the models but one thing to call out -- in CSharp we track all iDisposable resources, which includes user-created iDisposable-deriving classes (I had also crawled all of the iDisposable classes across .NET Framework/Core/Unity/Xamarin, etc so that we report on all of these when we come across them). One of the issues this brought up is that in the Java framework we were basing off of, having a model prevented the application of the user-defined constructor/dispose method for such classes, which is why we added model_with_analysis in PulseModelsCSharp.ml, used by both allocate_with_analysis and releaes_with_analysis.

infer/src/dotnet/ResourceLeaksCS.ml
matjin: Done

jvillard:
Thanks @archerD!

PR #1672, opened 2022-09-01
[java] fix the source and target compatibility failures on the Maven tests

jeremydubreil:
Compiling the Maven tests was failing before these changes with the following errors:
```
$ mvn --version
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 11.0.16, vendor: Ubuntu, runtime: /usr/lib/jvm/java-11-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.15.0-47-generic", arch: "amd64", family: "unix"

$ cd infer/infer/tests/build_systems/codetoanalyze/mvn/simple_app

$ mvn clean compile
...
[ERROR] Source option 5 is no longer supported. Use 7 or later.
[ERROR] Target option 5 is no longer supported. Use 7 or later.
...
```

Specifying the source and target version in the Maven configuration file fixes the errors and the compilation works with Java 8, Java 11 and Java 18.

This does not affect the Maven integration tests and all the tests are passing locally with `make -j test`.

### DIFF
--- a/infer/tests/build_systems/codetoanalyze/mvn/app_with_submodules/module2parent/module2/pom.xml
+++ b/infer/tests/build_systems/codetoanalyze/mvn/app_with_submodules/module2parent/module2/pom.xml
@@ -19,11 +19,8 @@
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
+          <encoding>UTF-8</encoding>
           <showWarnings>true</showWarnings>
-          <compilerArguments>
-            <Xlint:all/>
-            <Werror/>
-          </compilerArguments>
         </configuration>
       </plugin>
     </plugins>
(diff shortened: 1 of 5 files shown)


infer/tests/build_systems/codetoanalyze/mvn/app_with_submodules/module2parent/module2/pom.xml
```diff
@@ -19,11 +19,8 @@
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
+          <encoding>UTF-8</encoding>
           <showWarnings>true</showWarnings>
-          <compilerArguments>
-            <Xlint:all/>
-            <Werror/>
-          </compilerArguments>
```
jeremydubreil, line 26: Do we know why promoting warnings as errors is required here?

Without removing this, the compilation fails with:
```
[WARNING] bootstrap class path not set in conjunction with -source 8
```
However, it is not clear how to fix in the context of having the test compiling with different version of the Java compiler.

jvillard:
Welcome back @jeremydubreil!! :tada:

PR #1674, opened 2022-09-01
[java] fix missing dependency check in the Maven tests

jeremydubreil:
The variable `SOURCES` was not defined, which means that changing the Java sources would not trigger the test if those had already been run before. 

Without these changes, one could run:
```
$ make build_mvn_test
```
change the Java source file in a way that would change the list of Infer warnings, then, run again:
```
make build_mvn_test
```
and the test would still pass. This pull request fixes this.

All the tests are passing after the changes using `make -j test`.


jvillard:
Thank you!

PR #1675, opened 2022-09-02
[java] add support for Java WAR files

jeremydubreil:
Java WAR files are a special type of JAR file containing extra elements to deploy a full web applications: https://en.wikipedia.org/wiki/WAR_(file_format)

Infer should treat WAR file as JAR files and scan the bytecode they contain.


jvillard:
Thank you!

PR #1679, opened 2022-09-12
Test dependencies

jeremydubreil:
This pull request adds a test to document the `--dependencies` options triggering the analysis of the classes appearing in the Java classpath. This feature is especially useful when analysing JAR files packaging all the code of a Java application including the dependencies.

The examples added in the test are outlining how running the Pulse taint analysis with the dependencies mode enabled can eliminate cases of false negatives and cases of false positives.

All the tests are passing locally.

PR #1680, opened 2022-09-13
[java] use Temurin Java SDK 8 on GitHub Actions

jeremydubreil:
The Java tests should not depend on the platforms they are run on. However, they depend on the version of the Java SDK used to compiled the test code.

This pull request set a fixed version of Java on both platforms.


jvillard:
Sounds good, the tests pass on the GitHub action. Can this PR get out of draft mode?

jeremydubreil:
Thanks! The changes of this PR have been imported in https://github.com/facebook/infer/pull/1679 but I would rather have them merged independently so we should try to merge this one first.

jvillard:
I spoke too soon, the tests do not pass our internal CI, looks like we need to settle on Java 8 and not Java 11 for now.

PR #1681, opened 2022-09-23
[java] some simplifications and optimizations of the frontend

jeremydubreil:
This pull requests simplifies the code of the Java frontend to make it more efficient when running the capture on JAR files. On the two benchmarks I looked at, running `infer capture --generated-classes file.jar --sourcepath $PWD` was 1.6X faster after these changes in the first case and 2.2X faster in the second case.

These optimizations includes:
- using a hash table instead of a map to collect the list of java classes
- avoiding the repeated creation of strings when matching the packages
- avoid getting the classname from a node when the classname is already the key of the hash table
- changing the order of the source and package comparison to compare file names (cheaper) before comparing the packages (more expensive)

These changes are mostly intended to reduce the memory allocations when running the frontend on a JAR file. The benefit should be lower when running the capture with Buck but I don't expect the frontend to become slower either.


infer/src/java/jFrontend.ml
```diff
@@ -198,26 +198,18 @@ let create_icfg source_file program tenv icfg cn node =
 
 (* returns true for the set of classes that are selected to be translated in the given
    program *)
-let should_capture program package_opt source_basename node =
-  let classname = Javalib.get_name node in
-  let match_package pkg cn =
-    match JTransType.package_to_string (JBasics.cn_package cn) with
```
jeremydubreil, line 204: I noticed that the function `JTransType.package_to_string` is not used anywhere else so it can be deleted. I will update the PR.

davidpichardie: Hi Jeremy. I have started removing this deadcode myself. `JProcDesc.mem_classmap` was also dead now.

davidpichardie: I should be able to merge this PR soon. Thanks a lot @jeremydubreil!

davidpichardie:
Done! Merci beaucoup @jeremydubreil !

PR #1691, opened 2022-10-09
fix several typos about redundant words

x2018:
This PR fixes several typos about redundant words like `an an` and `the the` in documents or annotations.


x2018:
While reading the document, I unintentionally noticed a typo like these, so I further filtered them by several redundant word patterns.
I am not sure whether such changes are acceptable as they are relatively trivial, but they are indeed there.

jvillard:
Thank you @x2018!

PR #1709, opened 2022-12-08
BO: Add caller_procname to ModelEnv and remove proc_desc from OnDemandEnv.

dhauzar:
As suggested by @skcho:

"It could make more sense that adding caller's proc name to ModelEnv, rather than OndemandEnv (which is in Dom.Mem), since the ondemand env is the data for evaluating parameter values on-demand."

Proc_desc in OnDemandEnv is only used when creating a new Allocsite to file the procname of the caller of the function creating the Allocsite. This is needed in order to not end up with two identical allocsites created in two different subprograms by the call to the same builtin. In these situations, we either have already the correct procname (we don't need to get it from OnDemandEnv) or ModeEnv is available and we could add the procname to ModeEnv.

### DIFF
--- a/infer/src/bufferoverrun/bufferOverrunAnalysis.ml
+++ b/infer/src/bufferoverrun/bufferOverrunAnalysis.ml
@@ -311,8 +311,9 @@ module TransferFunctions = struct
         Dom.Mem.add_unknown ret ~location mem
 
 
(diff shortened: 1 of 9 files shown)


skcho:
Thank you, @dhauzar. Looks great!

dhauzar:
> Thank you, @dhauzar. Looks great!

You are welcome, thanks for upstreaming it!

PR #1722, opened 2023-01-11
Fingerprints in Sarif report

pkesseli:
Add `hash/v1` fingerprint to Sarif report.

### DIFF
--- a/infer/src/atd/sarifbug.atd
+++ b/infer/src/atd/sarifbug.atd
@@ -45,12 +45,18 @@ type codeflow = {
   threadFlows : threadflow list;
 }
 
+type fingerprints = {
+  hashV1 <json name="hash/v1"> : string;
+  key : string;
+}
+
 type sarifbug = {
   message : message;
   level : string;
   ruleId : string;
   ?codeFlows : codeflow list option;
   locations : loc list;
+  fingerprints : fingerprints;
 }
 
 type rule = {
--- a/infer/src/integration/SarifReport.ml
+++ b/infer/src/integration/SarifReport.ml
@@ -93,7 +93,8 @@ let loc_trace_to_sarifbug_record trace_list =
   List.map ~f:trace_item_to_record trace_list
 
 
-let pp_jsonbug fmt {Jsonbug_t.file; severity; bug_type; qualifier; line; column; bug_trace} =
+let pp_jsonbug fmt
+    {Jsonbug_t.file; severity; bug_type; qualifier; line; column; bug_trace; hash; key} =
   let message = {Sarifbug_j.text= qualifier} in
   let level = String.lowercase severity in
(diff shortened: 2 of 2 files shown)


infer/src/atd/sarifbug.atd
```diff
@@ -45,12 +45,17 @@ type codeflow = {
   threadFlows : threadflow list;
 }
 
+type fingerprints = {
+  hashV1 <json name="hash/v1"> : string;
```
geralt-encore, line 49: Any particular reason for it to be named `hashV1` instead of just `hash`?

jeremydubreil: In SARIF, [the fingerprints property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317654) allows to use different ways of computing issue identification. The idea is that whenever we want to use a different and incompatible hash mechanism here, we can just bump the version to `"hash/v2"`. However, I think it is probably fine to simply use `"hash"` for now and bump to `"hash/v1"` only once we need versioning.

We should probably also add the `key` field as part the fingerprints here. The key is expected to be stable whereas the `hash` could in theory depend on external things (i.e. the version of the compiler?). Adding the `key` here will be useful to compare findings over time, e.g. like:
```
   "fingerprints": {
      "hash" : "4ca6a6a9fe588dc9c25faac035d46f49",
      "key" : "Lists.java|removeInvalidatesNonEmptinessNPE|NULL_DEREFERENCE"
    }
```

pkesseli: According to Edoardo, who originally added this property in our own codebase, in order to be compliant with the SARIF spec it would have to be a version hierarchical string. [3.27.16](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541092), [3.5.4.2)](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540879).

Unless there is a strong reason to deviate from this, I'd suggest to start out with `hash/v1`. By the way, @geralt-encore, was your question with respect to the json `hash/v1` or just the ATD property name `hashV1`? I suppose for the latter it makes no difference, we could call it `hash` as well if preferred.

pkesseli: Updated the key property accordingly:
```json
"fingerprints": {
    "hash/v1": "71dd1e5127da495e4011c5acef571a0e",
    "key": "AnalysisStops.java|fieldReadInCalleeMayCauseFalseNegativeBad|NULLPTR_DEREFERENCE"
}
```

geralt-encore (approved):

PR #1723, opened 2023-01-18
Document `null` taint

pkesseli:
Add pulse test documenting `null` taint issue.


infer/tests/codetoanalyze/java/pulse/taint/NullPointer.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package codetoanalyze.java.pulse;
+
+class NullPointer {
+
+    void source(StringBuilder request) {
+    }
+
+    void foo() {
```
geralt-encore, line 15: Could you rename it to  something like`nullPointerTaintedFalsePositive` to make it more explicit?
Also, it would probably be nice to add a test case for String literals as well.

pkesseli: I renamed the methods and added an example for string literals.

geralt-encore (approved):
Thanks, I'll merge it as soon as CLA is sorted out

PR #1725, opened 2023-01-20
[java] prevent failing if same procedure is found in java bytecode

linpengzhang:
When people are running commands like `mvn compile` it will create .class files, and `mvn install` will create .jar files containing the class files already existing in the directory. This PR checks if the same procedure is found, by comparing its signature and if that's the case, it prevents Infer to fail.

### DIFF
--- a/infer/src/IR/Attributes.ml
+++ b/infer/src/IR/Attributes.ml
@@ -68,4 +68,8 @@ let load_from_uid, load, clear_cache, store =
   (load_from_uid, load, clear_cache, store)
 
 
+let load_formal_types pname =
+  match load pname with Some {formals} -> List.map formals ~f:snd3 | None -> []
+
+
 let is_no_return pname = match load pname with Some {is_no_return} -> is_no_return | _ -> false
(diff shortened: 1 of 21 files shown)


geralt-encore:
Hi Linpeng 🙂
Please, let me know if you have any issues with signing the CLA.

infer/src/IR/Cfg.ml
```diff
@@ -30,11 +30,19 @@ let get_all_defined_proc_names cfg =
...
+  || not (formals_eq old_attrs.formals new_attrs.formals)
+
+
 let create_proc_desc cfg (proc_attributes : ProcAttributes.t) =
   let pdesc = Procdesc.from_proc_attributes proc_attributes in
   let pname = proc_attributes.proc_name in
-  if Procname.Hash.mem cfg pname then
-    L.die InternalError "Creating two procdescs for the same procname." ;
+  if Procname.Hash.mem cfg pname && signature_eq pdesc (Procname.Hash.find cfg pname) then
+    L.die InternalError "Creating two procdescs with different signatures for the same procname." ;
```
jeremydubreil, line 45: I am not sure this will work. The backend will lookup the proc desc using the proc name only so adding more than one here is breaking the invariant expected by the backend.

However, the Java frontend also looks up the method by name following the Java class loading mechanism. So this case should not happen in theory and there may be a bug somewhere else that we would need to fix.

Also, we may want to hide the details of this in the CFG module. It should not be the responsibility of every frontend to make sure there is only one proc desc per proc name.

PR #1728, opened 2023-02-03
Bufferoverrun analysis - small fixes and improvements

dhauzar:
This PR contains small fixes and improvements to bufferoverrun analysis. They were originally implemented to fix the problems we discovered when using BO to analyze Ada code:

*1) Get also transitively reachable locations.*

Refactor get_reachable_locs_from_aux and get also transitively reachable
locations.

*2) Remove also frontend generated vars from the scope.*

The instruction ExitScope is generated for all variables
(both Var.LogicalVar and Var.ProgramVar) which become dead
according to the Liveness preanalysis.

BO implements this instruction by removing all variables of the
type Var.LogicalVar out of the abstract state and doing nothing
for variables of the type Var.ProgramVar. The rationale is that
the Liveness analysis doesn't take aliasing into account and
program variables can be aliased. That is, they can be accessed
after the instruction ExitScope through a pointer and removing
them from the abstract state is thus undesirable.

However, keeping all program variables in the abstract state
poses a performance problem - an example is an analysis of
a file gen_il-gen-gen_nodes.adb from GNAT sources. So this commit
removes the frontend generated variables from abstract states.
Such variables cannot be aliased so normally, it should be safe to
remove them. However, it seems that the content of these variables
is used by cost analysis so this is why this is put into a flag.
Perhaps, it would be safe to remove the variables also with
respect to the cost analysis when we are not in a loop.

*3) Improve pruning of simple aliases.*
    
Create pruning expression when pruning a simple alias equal to the pruned
logical variable.

*4) More precise handling of symbolic intervals.*

Handle symbolic intervals in more precise way. In particular (1) improve
the conversion of (symbolic) intervals to boolean values, and (2)
improve the equality of (symbolic) intervals. This is important to
make evaluation of expressions more precise.

* infer/src/bufferoverrun/itv.ml
(is_false): instead of comparing the interval with False (zero) directly,
return True if the interval is <= and >= zero. As the result, is_false
returns True also for [B(u), 0] and [max (I(s), 0), min (I(s), 0)].
(eq_sem): Return True also in the case when the bounds of two intervals
itv1 and itv2 are not equal and it holds that itv1 <= itv2 <= itv1.
(ne_sem): Use eq_sem instead of duplicating the logic of eq_sem.

*5) Fix importing of an alias bound to the return variable.*

The problem was that while an alias bound to a return variable was
always imported, it can be precisely imported only for locations
corresponding to global variables and locations reachable from formal
parameters passed by reference. Trying to import locations reachable from
formal parameters passed by value produced locations which are dereferences
of the locations that should be the correct result of the import.

Fix the problem by (1) fixing importing locations involving parameters
passed by value by importing them as unknown locations and (2) importing
only global locations and locations involving parameters passed by reference
when importing return value alias.

*6) Make the import of dereferences more robust.*

When importing a location representing a dereference from a callee
to a caller, fallback to attempting to create a symbolic location if
the target of the dereference is unknown in the caller.

Note the standard mechanism for creating symbolic locations can
fail because of the lack of type information. For example, in the case of
Ada, when calling a method on an abstract type that modifies a field of a
concretization of this type, we don't have type information for
the field in the caller.

Lift the condition on using the type passed to
BufferOverrunDomain.Valon_path when creating a value of a location
to always use this type when the type cannot be induced from formal,
actual parameters and type environment
(computed by BufferOverrunOnDemand.typ_of_param_path).
The rationale is that some instructions give type information
which cannot be computed by the mechanism above and it is too
bad to not use it. Since having type information means creating
a symbolic value, the consequence of this change can be that
a symbolic value can be created even for local variable. This is
no that bad since such value behaves as top and won't be imported
to callers of the current subprogram.


infer/src/bufferoverrun/bufferOverrunSemantics.ml
```diff
@@ -411,7 +411,24 @@ let rec eval_sympath_partial ~mode params p mem =
...
+            (let v = Mem.find loc mem in
+             if Val.is_default v then
+               (* There is no or default value for the dereference of a pointer in the abstract
+                  state of the caller. Create symbolic value representing the dereference.
+                  This makes it possible to import, e.g., the fact that "**A_Callee=2" for
+                  a pointer "*A_Caller" having an unknown value in the caller (A_Callee
+                  being a formal parameter of type pointer to pointer, A_Caller being an actual
+                  parameter of type pointer to pointer). *)
+               Mem.on_demand ~loc mem
+             else v )
```
skcho, line 428: * We may want to use `Mem.find_opt`?

* (If not) When the result of `Mem.find` is a default value, it is calling `Mem.on_demand ~loc mem`. But, IIUC, `Mem.find` already calls `Val.on_demand` in it (as below), and returns the `default` when the `on_demand` failed. So, I am confused about how calling `Mem.on_demand` again makes any difference.

https://github.com/facebook/infer/blob/51820faf5a553653559ce52845979153982c59e4/infer/src/bufferoverrun/bufferOverrunDomain.ml#L2171-L2176

infer/src/bufferoverrun/bufferOverrunDomain.mli
```diff
@@ -81,9 +81,16 @@ module Val : sig
 
   val of_literal_string : Typ.IntegerWidths.t -> string -> t
 
+  val of_ptr_loc : ?traces:BufferOverrunTrace.Set.t -> ptr_loc_type:Typ.t -> AbsLoc.Loc.t -> t
+  (** Create a value for a pointer pointing to a symbolic dereference of given location given the
+      type of the location. Return top if get_path returns None for loc (i.e., loc is LogicalVar or
+      Unknown or LiteralString Allocsite. *)
```
skcho, line 87: This function seems to be never called.

infer/src/bufferoverrun/bufferOverrunDomain.ml
```diff
@@ -2498,6 +2487,30 @@ module MemReach = struct
...
+                let locs =
+                  expand_reachable_locs
+                    ~locs:(LocSet.singleton (Loc.of_pvar pvar))
+                    ~expand_ptrs_arrs:false m
+                in
+                LocSet.fold (fun l m -> remove l (KeyLhs.of_loc l) m) locs m
+              else m )
+        else
+          Option.value_map (Var.get_ident var) ~default:m ~f:(fun id ->
+              remove (Loc.of_id id) (KeyLhs.of_id id) m ) )
```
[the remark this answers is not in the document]
dhauzar: Thanks, done.

infer/src/bufferoverrun/bufferOverrunSemantics.ml
```diff
@@ -416,12 +416,27 @@ let rec eval_sympath_partial ~mode params p mem =
...
+     not to actual expression. That is, if the actual expression is A, we can get its
+     value, not the location itself. This is also why we can import dereferences precisely
+     (we get powloc part of such value). *)
 and eval_locpath ~mode params p mem =
   let res =
     match p with
     | BoField.Prim (Symb.SymbolPath.Pvar _ | Symb.SymbolPath.Callsite _) ->
-        let v = eval_sympath_partial ~mode params p mem in
-        Val.get_all_locs v
+        PowLoc.unknown
```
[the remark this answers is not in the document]
dhauzar: Thank you for the question, the explanation I gave in the commit message was not clear and confusing. The fundamental problem fixed by this commit is indeed just the code in `eval_locpath` referenced above. Importing of alias of return value is only indirectly related to it (see below).

When `p` is `pvar` or `callsite` and we import its value by calling `eval_sympath_partial` then `Val.get_all_locs` returns locations representing possible dereferences of imported `p`, not locations representing imported `p`. On the other hand, when `p` is `deref`, importing the value of dereferenced pointer and calling `Val.get_all_locs` on the result gives locations representing imported `p` (this is the case below).

And indeed, there is no way how to get a location of a variable representing an actual parameter given a location representing a formal parameter as opposed to the case when we have a dereference of  formal parameter. This is because `BufferOverrunSemantics.ParamBindings` is a mapping from formals to evaluation of actuals so we can get a value that was passed to a call, but not the actual expression (from which we could get the variable if the expression is just a variable).

Now, the problem manifested itself when importing an alias bound to a return value, because normally a location representing a formal parameter itself is not imported (as opposed to its value or its dereference).

Here are examples when we cannot precisely import aliases of a return value and BO was importing them incorrectly before the fix:

```c
// alias of return value is x
int* callee_incorrectly_imported_alias (int* x) {
  return x;
}

int test_03 (int* y) {
  if (*(callee_incorrectly_imported_alias (y)) > 0) { // the alias of return value x from callee_incorrectly_imported_alias will be imported as *y
  ...
}

int test_04() {
  int i = 10;
  int* y = &i;
  if (*(callee_incorrectly_imported_alias (y)) > 0) { // the alias of return value x from callee_incorrectly_imported_alias will be imported as i
  ...
}
```

Here is the example when we cannot import an alias of return value precisely and BO was already correctly returning `Loc.unknown` before the fix:

```c
//  alias of return value is x
int callee_no_imported_alias (int x) {
  return x;
}

int test_01 (int y) {
  if (callee_no_imported_alias (y) > 0) {  // x was correctly imported as Loc.unknown
  ...
```

Here is the example, when we can precisely import an alias of return value:

```c
// alias of return value is *x
int callee_imported_alias (int* x) {
  return *x;
}

int test_02 (int y) {
  if (callee_imported_alias (&y) > 0) { // *x is correctly imported as y
```

infer/src/bufferoverrun/bufferOverrunSemantics.ml
dhauzar: Thanks for noticing this, I indeed messed up the code when implementing the switch to chose the default value (top / bottom).

The condition should be `if Val.top v then`. The interesting case is when the dereferenced pointer is present in the abstract state and its value is top. Then, `Val.on_demand` is not called in `Mem.find` and we would get `Loc.Unknown` as a location of its dereference. With the change, we call `Val.on_demand` and get a symbolic value representing the dereference. That is, it is supposed to handle an example like this:

```c
void callee (int** p) {
  **p = 2;
}

void test () {
  int* ptr;
  ptr = havoc; // ptr has unknown value
  callee (&ptr); // we want to import *ptr = 2
}
```

However, I decided to replace this code in the new commit "Make the import of dereferences more robust." that I added to this PR. The new code does basically the same thing but in a less obscure way and it is also more robust - in same cases, because of the lack of type information, `Val.on_demand` would return `Val.Unknown` resulting in returning and unknown location representing the imported dereference while we are able to create a symbolic location representing the dereference in the new code.

infer/src/bufferoverrun/bufferOverrunDomain.mli
dhauzar: Ah sorry, indeed, it is used only in our fork. I removed it.

skcho:
Ooops, sorry for the late review. I was waiting an email or a github notification, but somehow I seem to have messed my notification setting. I will get back to you soon.

infer/src/bufferoverrun/bufferOverrunSemantics.ml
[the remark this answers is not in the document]
skcho: Thank you so much for the explanation and great catch!! :clap::clap::clap::clap::clap:

infer/src/bufferoverrun/bufferOverrunSemantics.ml
skcho: It sounds great!

PR #1741, opened 2023-03-02
Stream-related tests

pkesseli:
Add Stream-related false negative tests.

- `copyBad`: works as expected
- `copyBadFN`: passing the `InputStream` to a method and reading from there leads to a false negative.
- `copyBad1FN`: Using `read`-style operations does not propagate the taint to the target array.
- `systemArrayCopyBadFN`: `System.arraycopy` does not propagate the taint to the target array.


infer/tests/codetoanalyze/java/pulse/taint/Streams.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.IOException;
+import java.io.File;
+
+public class Streams {
+
+  int bufferSize = 1024;
+
+  void copyBad() throws IOException {
+    InputStream tainted = (InputStream) InferTaint.inferSecretSource();
+    InferTaint.inferSensitiveSink(read(tainted.readAllBytes()));
```
geralt-encore, line 23: Seems like `readAllBytes` is available since java 9 but internally majority of our stuff is still on java 8 so it breaks the tests

pkesseli: For this example, I can try whether the issue is reproducible with just `toString` instead of `readAllBytes`. In general Infer's default propagation seems to handle `readAllBytes` correctly - the traditional `read` methods where the first argument should get tainted probably require some more attention.

pkesseli: Rebased the branch and replaced `readAllBytes()` by `toString().getBytes()`, which should be equivalent for taint analysis.

Issue #1750, opened 2023-04-06
Uninitialized false positive (normalization issue in PulseFormula?)


danielmercier:
In the following cpp reproducer:

```c++
#include <cassert>

void check_range(bool range_var_valid, int &range_var) {
  if (range_var_valid) {
    range_var = range_var + 1;
  }

  assert(range_var_valid >= 0 && range_var_valid <= 1);
}

void do_stuff(bool range_var_valid, int &range_var) {
  check_range(range_var_valid, range_var);
}

int main() {
  bool range_var_valid = false;
  int range_var;
  do_stuff(range_var_valid, range_var);
  return 0;
}
```

Pulse reports an uninitialized value when calling do_stuff for range_var. However, in case range_var_valid is false, range_var is not read.

Here is the summary for do_stuff (when range_var_valid is true):
```
conditions: {0 ≤ [v3]}∧{[a2] ≤ 1}∧{[a2] ≠ 0}
    phi: var_eqs: a2=v3
         && linear_eqs: a2 = -a4 +1 ∧ v5 = v6 -1
         && term_eqs: [v6 -1]=v5∧[-a4 +1]=a2
         && tableau: a2 = -a4 +1
         && intervals: a2=1
         && atoms: {is_int([v5]) = 1}∧{is_int([v6]) = 1}∧{is_int([a2]) = 1}
                   ∧{[a2] ≠ 0}
    { roots={ &range_var_valid=v1, &range_var=v2 };
      mem  ={ v1 -> { * -> v3 }, v2 -> { * -> v4 }, v4 -> { * -> v6 } };
      attrs={ v4 -> { WrittenTo (3, ) } };}
    PRE=[{ roots={ &range_var_valid=v1, &range_var=v2 };
           mem  ={ v1 -> { * -> v3 }, v2 -> { * -> v4 }, v4 -> { * -> v5 } };
           attrs={ v1 -> { MustBeInitialized(, t=0),
                           MustBeValid(, None, t=0) },
                   v2 -> { MustBeInitialized(, t=0),
                           MustBeValid(, None, t=0) },
                   v3 -> { UsedAsBranchCond(check_range, line 4, column 7, ) },
                   v4 -> { MustBeInitialized(, t=3),
                           MustBeValid(, None, t=3) } };}]
    need_specialization=false
    skipped_calls={ }
    Topl={len=0;content=
           [  ]}
```

Let's first zoom on the conditions part:
```
[...]

PR #1751, opened 2023-04-06
[pulse] More normalization in BaseDomain and Formula

danielmercier:
Normalize the conditions field from the path_condition according to the formula when normalizing it.

When canonicalizing the abductive state, also use var_repr for the pre state.

#1750


jvillard:
Thanks! The rest of pulse wasn't ready for this change at the time but now pulse normalises all abstract values on the fly, which in particular includes changes like this PR. Apologies for not incorporating this PR itself, that would have been tricky in the middle of all the changes. The particular false positive was fixed in 5d2d0c48aa9121a47c296d15da313c49404dc846. I believe we can close this.

PR #1752, opened 2023-04-11
[python] set version of Python to 3.8 on GitHub Actions

jeremydubreil:
This is already fixing the Linux CI job on GitHub. Still looking into why it does not work for macOS.


jeremydubreil:
@jvillard , Thanks! This would work. I was not able to fully look into why the paths are not the expected ones on macOS. There may be a way to fix this.

jvillard:
Fixed now, the GitHub builds are green again.

PR #1769, opened 2023-06-14
Add AdaCore to "Using Infer" list

yakobowski:


website/src/pages/index.js
```diff
@@ -53,6 +53,7 @@ const poweredByImages = [
   { url: "https://tangramflex.com", image: "/img/who/tangramflex.png" },
   { url: "https://www.uber.com", image: "/img/who/uber.svg" },
   { url: "https://www.whatsapp.com", image: "/img/who/whatsapp.svg" },
+  { url: "https://www.adacore.com", name: "/img/who/adacore.png" },
```
jvillard, line 56: this list is sorted alphabetically so AdaCore gets to be first

yakobowski: We feel so honored :bow:

yakobowski: Updated.

PR #1784, opened 2023-07-27
[java] Bump maven-compiler-plugin version in capture profile

### LINKED ISSUE #990: infer causing Maven 3 version missing warning for org.apache.maven.plugins:maven-compiler-plugin
I am getting the following warning when using `infer` in maven mode... 

> Capturing in maven mode...
> [INFO] Scanning for projects...
> [WARNING]
> [WARNING] Some problems were encountered while building the effective model for com.shpsec:pegasus-build:pom:5.0-SNAPSHOT
> [WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-compiler-plugin is missing. @ line 75, column 19
> [WARNING]
> [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
> [WARNING]
> [WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
> [WARNING]

this seems to be an new requirement in Maven 3 that version numbers are specified for plugins...

`infer` is appending the following `<profile>` to the `pom.xml` file... 
```xml
<profiles><profile>                                                                                       
      <id>infer-capture</id>                                                                              
      <build>                                                                                             
        <plugins>                                                                                         
          <plugin>                                                                                        
            <groupId>org.apache.maven.plugins</groupId>                                                   
            <artifactId>maven-compiler-plugin</artifactId>                                                
            <configuration>                                                                               
              <compilerId>javac</compilerId>                                                              
              <forceJavacCompilerUse>true</forceJavacCompilerUse>                                         
              <fork>true</fork>                                                                           
              <executable>/usr/local/Cellar/infer/0.15.0/lib/infer/infer/bin/infer</executable>           
            </configuration>                                                                              
          </plugin>                                                                                       
        </plugins>                                                                                        
      </build>                                                                                            
    </profile></profiles></project>  
```

kohlschuetter:
The "infer-capture" Maven profile specified a version for maven-compiler-plugin to fix #990.  This may interfere with the inspected project's configuration.

In this particular case, version 3.8.0, was specified. This prevented compilation of some projects, resulting in the ominous error 'Cannot invoke "java.lang.Throwable.getCause()" because "cause" is null'.

Removing the declared version for maven-compiler-plugin is a potential fix, which would automatically use the version defined in the origin Maven project. However that may break Maven 2.x projects because "forceJavacCompilerUse" is only available since version 3.0.

Bumping version to the latest available 3.11.0 instead.


geralt-encore (approved):
Thanks for the fix!

PR #1820, opened 2024-03-20
Minor grammatical edits and typo fixes

kennyballou:
Just a small set of edits for separation logic and biabduction.

### DIFF
--- a/website/docs/02-separation-logic-and-biabduction.md
+++ b/website/docs/02-separation-logic-and-biabduction.md
@@ -21,7 +21,7 @@ composing the reasoning chunks together.
 Separation logic is based on a logical connective <Math code="\\( * \\)" />
 called the _separating conjunction_ and pronounced "and separately". Separation
 logic formulae are interpreted over program allocated heaps. The logical formula
-<Math code="\\( A*B \\)" /> holds of a piece of program heap (a heaplet) when it
+<Math code="\\( A*B \\)" /> holds if a piece of program heap (a heaplet) when it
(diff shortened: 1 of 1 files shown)


website/docs/02-separation-logic-and-biabduction.md
```diff
@@ -21,7 +21,7 @@ composing the reasoning chunks together.
 Separation logic is based on a logical connective <Math code="\\( * \\)" />
 called the _separating conjunction_ and pronounced "and separately". Separation
 logic formulae are interpreted over program allocated heaps. The logical formula
-<Math code="\\( A*B \\)" /> holds of a piece of program heap (a heaplet) when it
+<Math code="\\( A*B \\)" /> holds if a piece of program heap (a heaplet) when it
```
jvillard, line 24: I believe the original text was correct here and we can say "The formula holds *of* a piece of heap when ...". "if" is incorrect here.

website/docs/02-separation-logic-and-biabduction.md
```diff
@@ -47,7 +47,7 @@ formula above can be visualized like so:
 
 ![](/img/SepSplit.jpg)
 
-The important thing about separating conjunction is the way that it fits
+The important thing about separating conjunctions is the way that it fits
```
jvillard, line 50: The singular was correct here (referred to by "it" right after). Maybe "The important thing about *the* separating conjunction is ..."?

jvillard (commented):
Thanks for the fixes. There are two places that I think should be changed back or changed differently but I can do that when importing the PR if you don't mind.

PR #1825, opened 2024-03-27
Fix the ocamlformat install step in the GitHub worflow

jeremydubreil:
The code was previously using `$INFER_ROOT` which is not defined for this workflow.


jvillard:
Thanks for fixing this!

PR #1938, opened 2025-08-23
fix clang build with glibc 2.42 re: struct termio

etwoo:
I hit some issues building Infer from source in my Arch Linux dev environment. The problem seems to be specific to compiling against glibc 2.42, which Arch [recently accepted into its official repositories](https://gitlab.archlinux.org/archlinux/packaging/packages/glibc/-/commit/3543edc0a9e583f9c3880c197d83409c5e8d4020).

The fix in this PR is to backport fixes from LLVM upstream for https://github.com/llvm/llvm-project/issues/137321

- https://github.com/llvm/llvm-project/commit/59978b2
- https://github.com/llvm/llvm-project/commit/c99b1bc

### DIFF
--- a/facebook-clang-plugins/clang/src/glibc_2.42_struct_termio_interceptors.patch
+++ b/facebook-clang-plugins/clang/src/glibc_2.42_struct_termio_interceptors.patch
@@ -1,0 +1,64 @@
+From 59978b21ad9c65276ee8e14f26759691b8a65763 Mon Sep 17 00:00:00 2001
+From: Tom Stellard <tstellar@redhat.com>
+Date: Mon, 28 Apr 2025 13:45:11 -0700
+Subject: [PATCH] [sanitizer_common] Remove interceptors for deprecated struct
+ termio (#137403)
+
+This struct will be removed from glibc-2.42 and has been deprecated for
+a very long time.
+
+Fixes #137321
+---
+ .../sanitizer_common_interceptors_ioctl.inc               | 8 --------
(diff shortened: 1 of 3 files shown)


etwoo:
Assuming the feedback from https://github.com/facebook/infer/pull/1939#issuecomment-3297464996 would also apply here, I've rebased this bugfix branch onto HEAD of main.

ngorogiannis (approved):

etwoo:
Hi @ngorogiannis, I see an [import notification for this PR](https://github.com/facebook/infer/pull/1938#issuecomment-3995780621) from March 3, but there does not seem to be a subsequent action from @meta-codesync to close this PR and add the "Merged" label. Did something go wrong after import, and if so, is there anything I can do to correct it?

Edit: is this PR perhaps no longer necessary, assuming infer has upgraded its toolchain to an upstream LLVM release that already includes this fix, like in e09d41e?

etwoo:
> Edit: is this PR perhaps no longer necessary, assuming infer has upgraded its toolchain to an upstream LLVM release that already includes this fix, like in [e09d41e](https://github.com/facebook/infer/commit/e09d41e67fc4b14da7deeff59364247cba66db38)?

Confirmed -- I created a fresh Arch Linux VM with the latest glibc (now 2.43), pulled the latest infer repo, and built from source. The patches in this PR no longer appear to be necessary (perhaps as a result of changes like e09d41e, 86c3c12).

Closing this PR accordingly.

PR #1939, opened 2025-08-23
ignore suppressed issues for process exit code

etwoo:
In my usage, `infer run` fails with a non-zero exit code even after I have suppressed all issues in my C project with `@infer-ignore` comments (along with appropriate `.inferconfig` settings).

I tracked this down to how `fail_on_issue_epilogue()` maps presence/absence of issues to the overall process exit code. The fix in this PR is to make Infer only return a non-zero exit code if one or more _unsuppressed_ issues exist. In other words, Infer should treat a suppressed issue as equivalent to an entirely non-existent issue, for the purposes of the overall `infer run` exit code.

Please let me know if there are backwards-compatibility concerns that would require this behavior to be gated by an opt-in CLI argument. I am hoping it is acceptable to change the default behavior in this manner, in order to follow the principle of least surprise for people (like myself) using Infer in their CI pipelines.

I've added a test for this behavior, based on the existing `fail_on_issue` testcase. I hope the latter is an acceptable template for a new test; please let me know if a different testcase would make for a better starting point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ BUILD_SYSTEMS_TESTS += \
   infer-debug \
   j1 \
   missing_deps \
+  pass_on_suppression \
   procs_to_analyze \
   project_root_rel \
   pulse_messages_c \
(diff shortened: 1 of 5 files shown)


martintrojer:
@etwoo can you please rebase (to improve the ci signal somewhat)?

PR #1940, opened 2025-08-24
Fix Get Infer command in 00-getting-started.md

mikoto2000:
I was unable to install even after running the commands as in Getting Started.
This fixes that problem.


davidpichardie:
Please reopen a PR if needed.

PR #1944, opened 2025-09-11
Pulse-infinite, the new INFINITE_LOOP checker in Pulse

jvanegue:
Pulse-infinite, the new INFINITE_LOOP checker in Pulse

Tool Behind paper: "Non-Termination Proving: 100M LoC and Beyond" https://www.arxiv.org/abs/2509.05293


infer/src/pulse/PulseExecutionDomain.ml
```diff
@@ -98,10 +105,220 @@ let pp_with_kind kind path_opt fmt exec_state =
...
+    in
+    let rec find_duplicate_state (lst : t list) (curi : int) (maxi : int) : int =
+      match lst with
+      | hd :: tl ->
+          let isrep = is_repeated hd tl in
+          (* L.debug Analysis Quiet "JV: Found duplicate state in POST with index = %d \n" curi; *)
+          if isrep && curi > maxi then find_duplicate_state tl (curi + 1) curi
+          else find_duplicate_state tl (curi + 1) maxi
+      | _ ->
+          (* L.debug Analysis Quiet "JV: Chosen duplicate in POST has index = %d \n" maxi; *) maxi
```
davidpichardie, line 303: Hi @jvanegue . Do you expect this case to be feasible? 
As far as I understand, you will return -1 in this situation and then you will end up in line 280 above.

jvanegue: Good question - I think you're right. I am not sure we can ever reach this on any particular codebase. If it is, the desired behavior is encoded in this case, which is to not add any infinite state to minimize the number of alerts. There are (disabled) debug lines near line 280 and 317 which can be enabled to see this ever happen in the logs.

PR #1975, opened 2025-11-30
Fix SARIF reports to comply with specification

### LINKED ISSUE #1974: SARIF reports don't comply with the specification
According to the [SARIF specification](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790677), section 3.4.4
> If this artifactLocation object describes a top-level artifact and the value of its uri property (§3.4.3) is a relative reference, the artifactLocation object SHOULD contain a property named uriBaseId whose value is a string which indirectly specifies the absolute URI with respect to which that relative reference is interpreted. If the uri property contains an absolute URI, the uriBaseId property SHALL be absent. If this artifactLocation object describes a nested artifact, uriBaseId SHALL be absent.
> 
> If a SARIF consumer requires an absolute URI (for example, to display the specified artifact to a user), then it needs to resolve uriBaseId to an absolute URI, which it can then combine with the relative reference stored in the uri property.

Currently, the `uriBaseId` field contains an absolute path to the source file, starting with `file:`, and the `uri` contains a relative path starting with `file:`, however this is not intellegible by standard SARIF consumers.

Solving the issue is very simple, as it entails either
- Removing the the `file:` prefix from the `uri` field, and leaving it as is otherwise (so the relative path to the file without any prefix) and **either**:
  - remove the `uriBaseId` property, which would lead to many tools just assuming the project root to be the base folder, leading to the sarif reports also being portable (by not referencing any absolute path); after all the property is not required by the spec
  - put the absolute path of the **project root**, prepending `file:` and without appending the relative path of the file
- Putting the absolute path of the file, prepended by `file:`, as the `uri` property, and omitting the `uriBaseId` property, as required by the spec; _this seems the worst solution of the bunch, as it is the hardest to port between systems, while a text replace with vim or cli tooling is totally doable in any of the cases_

----

I will open a pr going with the second option (relative paths in `uri` and absolute path of the project root in `uriBaseId` as it seems to be the one most in line with the original intentions of the code.

friedbyalice:
Fixes #1974

_Please note this is a personal contribution, not on behalf of my employer_

### DIFF
--- a/infer/src/integration/SarifReport.ml
+++ b/infer/src/integration/SarifReport.ml
@@ -69,9 +69,8 @@ let pp_results_header fmt =
 
 let loc_trace_to_sarifbug_record trace_list =
   let file_loc filename =
-    let absolute_source_name = Config.project_root ^/ filename in
-    let file_path = "file:" ^ filename in
-    {Sarifbug_j.uri= file_path; Sarifbug_j.uriBaseId= absolute_source_name}
+    let absolute_project_root = "file:" ^ Config.project_root in
+    {Sarifbug_j.uri= filename; Sarifbug_j.uriBaseId= absolute_project_root}
   in
   let message description = {Sarifbug_j.text= description} in
   let region line_number column_number =
@@ -102,9 +101,8 @@ let pp_jsonbug fmt
   in
   let level = String.lowercase severity in
   let ruleId = bug_type in
-  let absolute_source_name = Config.project_root ^/ file in
-  let file_path = "file:" ^ file in
-  let file_loc = {Sarifbug_j.uri= file_path; uriBaseId= absolute_source_name} in
+  let absolute_project_root = "file:" ^ Config.project_root in
+  let file_loc = {Sarifbug_j.uri= file; uriBaseId= absolute_project_root} in
   let region =
     match column with
     | -1 ->


infer/src/integration/SarifReport.ml
```diff
@@ -69,9 +69,8 @@ let pp_results_header fmt =
 
 let loc_trace_to_sarifbug_record trace_list =
   let file_loc filename =
-    let absolute_source_name = Config.project_root ^/ filename in
-    let file_path = "file:" ^ filename in
-    {Sarifbug_j.uri= file_path; Sarifbug_j.uriBaseId= absolute_source_name}
+    let absolute_project_root = "file:" ^ Config.project_root in
+    {Sarifbug_j.uri= filename; Sarifbug_j.uriBaseId= absolute_project_root}
```
jeremydubreil: Alternatively, `uriBaseId` can be set to the constant `"%srcroot%"` as described [here](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790681). This helps when scanning repositories where we typically want the SARIF report to be the same (for the same commit) regardless of where the local checkout appears on the filesystem.

dulmarod:
Please rebase and resubmit if this is still required.

friedbyalice:
Superseded by #2077

PR #2017, opened 2026-03-06
[rust] Heap types (boxes)

ArjanSeijs:
- Adds support for boxes for primitives types.
-  Adds a new builtin function boxnew with a sil-model  `lib/rust/models.sil` that allocates data on the heap and  returns a pointer
    -  The `lib/rust/models.sil`  file also contains a explantion of how boxes are treated by various stages of ULLBC.
- Since rust type system is quite elaborate with heap allocated types support for adding charon arguments is added to remove some of the boiler plate with arguments like --hide-marker-traits and --hide-allocator.
- Additionally we give a warning if we encounter type definitions of functions that cannot be translated instead of dying so that it is still possible to analyze the functions we do support.


dulmarod:
@ArjanSeijs  be aware that the biabduction checker has been removed from the codebase, so the option `--enable-issue-type BIABDUCTION_MEMORY_LEAK` is not needed, I can remove it before shipping the PR. When looking at infer's documentation, use the next version of the website https://fbinfer.com/docs/next/getting-started, that's the latest because we haven't made a new release in a while.

PR #2018, opened 2026-03-08
Fix Pulse NULL_DEREFERENCE false negative caused by irrelevant inequa…

bhuvii2005:
Title: Fix Pulse NULL_DEREFERENCE false negative caused by irrelevant inequalities

(Note: This contribution was developed with the assistance of an AI coding agent to trace the OCaml bug through the Pulse arithmetic logic.)

Summary: Infer's Pulse analysis currently fails to report a NULL_DEREFERENCE when the null-dereference occurs in a branch that is preceded by an irrelevant if-else statement checking a constant non-null value (e.g., if (System.out != null)).

This happens because PulseFormula.is_manifest incorrectly suppresses the issue as a "latent" issue. When evaluating the condition System.out != null, Pulse adds an inequality constraint [System.out ≠ 0] to the path condition. is_manifest then evaluates this constraint—which involves an unallocated variable—and assumes the path is only possible in specific caller contexts, thereby suppressing the NULL_DEREFERENCE that occurs later on a completely unrelated variable.

Changes: This PR makes PulseFormula.is_manifest context-aware by piping down the diagnostic parameter from LatentIssue.should_report.

Extracted the invalid address variable from the AccessToInvalidAddress diagnostic.
Updated is_manifest to ignore neq_zero conditions if the variable involved in the constraint is unrelated to the variable that was null-dereferenced.
Piped ~diagnostic through PulseLatentIssue.should_report, PulseArithmetic, and down to PulseFormula.
Test Plan: 
Added test1_useless_branch_npe_Bad and test2_useless_branch_npe_Ok to infer/tests/codetoanalyze/java/pulse/NullPointerExceptions.java.

 This guarantees that the NULL_DEREFERENCE is correctly caught on the array variable even when preceded by an irrelevant System.out == null check.

### DIFF
--- a/infer/src/pulse/PulseArithmetic.ml
+++ b/infer/src/pulse/PulseArithmetic.ml
@@ -121,8 +121,8 @@ let prune_eq_one v astate =
 
 let is_known_zero astate v = Formula.is_known_zero astate.AbductiveDomain.path_condition v
 
-let is_manifest summary =
-  Formula.is_manifest (AbductiveDomain.Summary.get_path_condition summary) ~is_allocated:(fun v ->
+let is_manifest ?diagnostic summary =
+  Formula.is_manifest ?diagnostic (AbductiveDomain.Summary.get_path_condition summary) ~is_allocated:(fun v ->
       AbductiveDomain.Summary.is_heap_allocated summary v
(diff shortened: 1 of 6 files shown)


davidpichardie:
Closing. High risk of FP.

PR #2035, opened 2026-04-15
[rust] [1/2] Update Charon to version 0.1.177 to be able to use 

ArjanSeijs:
`--precise-drops`. This also sets the mir to elaborated and changes how boxes are represented at the mir level.

Box support will be updated  in the next commit.


dulmarod:
Looks like there is a build failure, you could try running it again.

PR #2036, opened 2026-04-15
[rust] [2/2] Support the elaborated box reprenstation for precise-drops 

ArjanSeijs:
and some partial support for modeling drops when retrieving drop glue with --desugar-drops.

We do some abstraction over Box, and NonNull since the compiler has some special handling for how it treats Box and NonNull and their ABI are equivalent.

Merge after #2035


dulmarod:
@ArjanSeijs please can you rebase this PR so that we can try to import it? The PR 1 is already in master.

PR #2051, opened 2026-06-18
minor changes around pulse's solver and eternal checker

jvillard:
Some preparatory work to improvements to Pulse's SMT solver and its Eternal infinite loop checker. See individual commits for easier review.

- add two missing .mli files
- minor changes to eternal algorithms (paging @davidpichardie)
- small refactors in Pulse's SMT solver

PR #2052, opened 2026-06-18
[eternal][NEW] check that the loop invariant is satisfied by the entry state

jvillard:
[Ideally to be merged on top of #2051 to preserve some of the history.]

Whenever pulse-eternal finds an infinite loop, check that the invariant found is satisfied by the abstract state we had at the loop entry. This is required for soundness (avoiding FPs).

This adds an implication checker to pulse's solver, although it makes some assumptions about the shape of the abstract states because it is only used for eternal.


dulmarod:
ERROR 'Running test: direct erlang pulse test'

PR #2053, opened 2026-06-18
improvements to pulse's solver

jvillard:
Three main changes to pulse's SMT solver:
1. Avoid infinite loops in the solver due to propagating new facts to the rest of the formula. This is done by being more systematic about fuel accounting.
2. Fix a few places where not enough normalisation was done by the solver, although more work is needed here to make sure that the formula is always in normal form. This improves completeness (of the solver) and avoids running out of fuel.
3. teach pulse a few things about modulo arithmetic to handle simple examples of non-termination where parity abstraction is needed (new infinite loops found in examples)

I have only updated the C and C++ tests so other test changes might need to be recorded (I can update the PR once I see the CI results).


dulmarod:
ERROR 'Running test: direct hack pulse test'. @jvillard  that one may be harder for you to fix I expect. I can help if needed.

jvillard:
@dulmarod I don't have hackc set up, could you record the new test results if they look ok please?

jvillard:
I forgot to mention but this PR goes on top of #2052 if possible.

PR #2054, opened 2026-06-20
[pyml][ppx_show] remove unneeded pins

ngorogiannis:
These are no longer necessary as the stdcompat dependency has been updated upstream.

Some changes were required as a few packages were upgraded, and also because some python unit tests broke.

Also fixed a preexisting bug related to the differences between linux and OSX sed.


ngorogiannis:
Hey @davidpichardie, my aim was to ease the upgrade to OCaml 5.4.1, but unpinning ppx_show and pyml is still a blocker, so best not land this PR in favour of #2056.

PR #2059, opened 2026-06-28
[rust][TreeBorrows] Pointer reference and mutability information

MateiMirica:
First step towards a Tree Borrows checker for the Rust frontend.

Charon already distinguishes references (&/&mut) from raw pointers (*const/*mut), and mutable from const, but that information was discarded during translation. This PR preserves it: pointer types now carry .rust_pointer and .rust_mut attributes in Textual, and the Textual→SIL bridge maps them onto the SIL type_quals (is_reference, is_const), so later analyses can tell a &mut from a *mut from a &.

This is a frontend translation only,  no analysis changes.

### DIFF
--- a/infer/src/IR/Typ.mli
+++ b/infer/src/IR/Typ.mli
@@ -64,6 +64,8 @@ val mk_type_quals :
 
 val is_const : type_quals -> bool
 
+val is_reference_on_source : type_quals -> bool
+
 val is_restrict : type_quals -> bool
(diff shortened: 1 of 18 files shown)


dulmarod:
The CI is showing tests errors. ERROR 'Running test: direct rust sil test'

dulmarod:
Can you try rebasing to fix the build?

PR #2060, opened 2026-06-28
[rust][TreeBorrows] Register the Tree Borrows checker

MateiMirica:
Registers the Tree Borrows checker and the TREE_BORROWS_UB issue type.


dulmarod:
The tests are failing because you need to run `make manuals` after adding a new checker. Then add the changed files to the diff.

dulmarod:
See CI failures, the formatting seems to be failing. `make fmt` should fix that.

PR #2062, opened 2026-06-29
[multicore] fix bug in compaction threshold

ngorogiannis:
Calculation uses the word size in bits, but doesn't multiply the number of bytes by 8.


dulmarod:
Looks like there are formatting issues.

PR #2065, opened 2026-07-02
[semdiff] reduce B006 migration false positives

davidpichardie:
## Summary

Reduces false positives (and eliminates crashes/timeouts) in the Python **B006**
migration checker (`infer semdiff --semdiff-b006-migration`), which verifies that a
mutable-default-argument codemod (`def f(x=[])` → `def f(x=None): if x is None: x=[]`)
preserves behaviour. Builds on #2063.

Evaluated on **796 real B006 codemod diffs**: the false-positive rate drops from **~55%** (baseline) to **~15.6%**, with **no new false negatives**, **no crashes**, and **no timeouts**. Each fix is verdict-preserving for equivalence and keeps the check sound (an inverted `is None` guard, or a changed value, still diverges).

## The fixes (one per commit)

1. **Skip class bodies; recover method/nested-function defaults.** A method's default
   is baked into the class-body proc (not the skipped `__module_body__`); it was
   compared and diverged on every migrated method. Skip class bodies like the module
   body, and recover defaults from the enclosing scope wherever it is.
2. **Model empty container constructors as pure.** `set()`/`dict()`/`list()`/`tuple()`
   ≡ the empty literal; canonicalise to the pure empty-container builtin.
3. **Recover elements of large (≥3) list/set/dict literals.** CPython builds these
   with `BUILD_x 0` + in-place `EXTEND`/`UPDATE`; recover the elements so the literal's
   value is complete on both sides.
4. **Fix two exponential blowups on large functions.** (a) callee-name lookup printed
   the whole callee atom (expanding the state DAG) — cap the print depth; (b) the
   bisimulation had no memo — add a sound positive-result memo. Removes all timeouts
   (a 49s conversion → 0.01s).
5. **Handle keyword-only parameters (and unseeded locals).** `python_args` lists only
   positional (`co_argcount`) params, so keyword-only params crashed with
   `py_load_fast: variable not found`. Recover keyword-only defaults from
   `py_make_function`'s kwdefaults, and fall back to an opaque local instead of
   aborting the file.
6. **Treat `dict(k=v, …)` as the literal `{k: v, …}`.** The common mmcv/mmdet
   `dict(type=…)` config style; canonicalise the keyword dict constructor to the same
   `py_build_map` the literal produces.
7. **Same `dict(k=v)` canonicalization in the statement path**, for tuple/nested
   defaults that keep `dict(…)` in the guard body.

## Test plan

- New `semdiff-b006` regression tests for every fix (EQUAL for faithful migrations,
  DIFFERENT for inverted guards / changed values).
- Existing `semdiff` unit + e2e suites unchanged and passing.
- Large-corpus evaluation (796 real diffs): FP ~55% → ~15.6%, 0 FN, 0 crash, 0 timeout.

The remaining false positives are dominated by closure-captured (cell-variable)
defaults and module-vs-function name-resolution scope differences, which need
soundness-sensitive modelling changes and are left as (conservative) future work.

### DIFF
--- a/infer/src/semdiff/Semdiff.ml
+++ b/infer/src/semdiff/Semdiff.ml
@@ -92,10 +92,15 @@ let proc_structurally_equal proc_old proc_new =
 let semdiff_b007_textual ~debug (module_old : Textual.Module.t) (module_new : Textual.Module.t) =
   let procs_old = extract_procs module_old in
   let procs_new = extract_procs module_new in
+  let class_bodies = StructuredPeg.find_class_body_procs module_old in
+  let is_class_body proc =
+    List.mem class_bodies proc.Textual.ProcDesc.procdecl.qualified_name
+      ~equal:Textual.QualifiedProcName.equal
+  in
   let all_accepted =
     List.for_all procs_old ~f:(fun proc_old ->
         let name = proc_fun_name proc_old in
-        if is_module_body name then true
+        if is_module_body name || is_class_body proc_old then true
         else
           match List.find procs_new ~f:(fun p -> String.equal (proc_fun_name p) name) with
           | None ->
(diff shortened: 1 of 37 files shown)


dulmarod:
Looks like formatting needs fixing.

Issue #1951, opened 2025-10-13
False Negative in PULSE_RESOURCE_LEAK Detection


Zustin:
Hi, the following code example contains a false negative bug. Infer failed to report a `PULSE_RESOURCE_LEAK` warning at line 10, where an `InputStream` resource is not closed after use. The infer version is v1.2.0.

```java
import java.io.FileInputStream;
import java.io.FileNotFoundException;
import java.io.IOException;
import java.util.Optional;

public class Main {
    void foo(boolean b) throws IOException, FileNotFoundException {
        Optional<FileInputStream> stream;
        if (b) {
            stream = Optional.of(new FileInputStream("file.txt")); // should report a warning at this line
        }
    }
}
```

**Log**
```bash
infer run -- javac Main.java
Capturing in javac mode...
Found 1 source file to analyze in /infer-out
0/3 [................................................................................] 0% 120ms
⊢ [ 0.1s][17.8M] Main.java: Main.<init>()
⊢ [ 0.1s][17.8M] Main.java: void Main.foo(boolean)
3/3 [################################################################################] 100% 170ms
⊢ [ 0.0s][17.8M] idle
⊢ [ 0.0s][17.8M] idle
⊢ [ 0.0s][17.8M] idle

  No issues found  
```

SuryanshSS1011:
Hey @davidpichardie, I've been looking into this and would like to take it.

Following your suggestion I started from the `java.lang.Boolean` model in `PulseModelsJava.ml`. While scoping it I think the precise shape here may be a bit different from `Boolean`, so I wanted to check the approach with you before opening a PR.

`Boolean` is a value box (backing field + `init`/`equals`/`valueOf`/`booleanValue`). But `Optional.of(new FileInputStream(...))` is really a wrapper around a resource, so the leak obligation already exists on the underlying stream from its `<init>`. The gap is that `Optional` is unmodeled, so closing through it (or never closing it) isn't tracked.

My plan, as an alternative to a plain value-box model is to:

- Add an `Optional` module with an internal backing field (`__infer_model_backing_optional`), modeling `of` / `ofNullable` to store the argument (mirroring `Boolean.init`'s `store_field`), `get` to load it back, and `empty` for the empty case.
- Wire matchers with `+map_context_tenv (PatternMatch.Java.implements "java.util.Optional")`, alongside the `Boolean` rows.

[...]

PR #2068, opened 2026-07-02
[infer][pulse] model java.util.Optional to catch resource leaks through it

SuryanshSS1011:
## Summary

Pulse missed `PULSE_RESOURCE_LEAK` when a resource (e.g. a `FileInputStream`) is wrapped in a `java.util.Optional` and never closed, because `Optional` was unmodeled. This adds a small model for `Optional` so the obligation on the wrapped resource is tracked through it.

Fixes the false negative reported in #1951:

```java
void foo(boolean b) throws IOException {
    Optional<FileInputStream> stream;
    if (b) {
        stream = Optional.of(new FileInputStream("file.txt")); // now reported
    }
}
```

## Changes

- `PulseModelsJava.ml`: new `Optional` module modeling `of` / `ofNullable` (store the wrapped value into a backing field) and `get` (load it back), mirroring the existing `Boolean` model. The wrapped resource keeps the `JavaResource` obligation from its own constructor, so `optional.get().close()` releases it and an `Optional` whose contents are never closed reports a leak.
- `ResourceLeaks.java`: tests for the leaking (`of`, `ofNullable`) and the closed-via-`get()` cases.
- `issues.exp`: the two new expected reports.

## Testing

`make -C infer/tests/codetoanalyze/java/pulse test` passes. Verified against a stock build at the same commit: the repro reports no issue before the change and `PULSE_RESOURCE_LEAK` after, with `optionalOfClosedOk` staying clean.


SuryanshSS1011:
@davidpichardie

This implements the Optional model we discussed in #1951 (of/ofNullable/get), mirroring the Boolean model. Verified the repro reports PULSE_RESOURCE_LEAK after the change and stays clean when closed via get().

Issue #1937, opened 2025-08-22
variable argument list pointers aren't traced as allocation targets resulting in false positives for PULSE_UNINITIALIZED_VALUE and MEMORY_LEAK_C


grooverdan:
The use of a variable argument list to allocate a memory block once for its argument pointers at once isn't traced to the returning function.

x.c:
```c
#include <stdarg.h>
#include <stdlib.h>

void *ma_multi_malloc(int myFlags, ...)
{
  va_list args;
  char **ptr,*start,*res;
  size_t tot_length,length;

  va_start(args,myFlags);
  tot_length=0;
  while (va_arg(args, char **))
  {
    length=va_arg(args, size_t);
    tot_length+=length;
  }
  va_end(args);
         
  if (!(start=(char *)malloc(tot_length)))
    return 0;
      
  va_start(args,myFlags);
  res=start;
  while ((ptr=va_arg(args, char **)))
  {
    *ptr=res;
    length=va_arg(args,size_t);
    res+=length;
  }
  va_end(args);
  return start;
}

void test()
{
  char *v, *u; 
  ma_multi_malloc(0, &v, (size_t) 1, &u, (size_t) 3);
  *v='d';
  *u='o';
  free(v);  
}
```

```
3/3 [################################################################################] 100% 52.5ms

/tmp/x.c:37: error: Memory Leak(MEMORY_LEAK_C)
  Memory dynamically allocated by `malloc`, indirectly via call to `ma_multi_malloc()` on line 37 is not freed after the last access at line 37, column 3. 
  35. {
  36.   char *v, *u;
  37.   ma_multi_malloc(0, &v, (size_t) 1, &u, (size_t) 3);
        ^
  38.   *v='d';
  39.   *u='o';

/tmp/x.c:38: error: Uninitialized Value(PULSE_UNINITIALIZED_VALUE)
  `v` is read without initialization. 
  36.   char *v, *u;
  37.   ma_multi_malloc(0, &v, (size_t) 1, &u, (size_t) 3);
  38.   *v='d';
         ^
  39.   *u='o';
  40.   free(v);  

/tmp/x.c:39: error: Uninitialized Value(PULSE_UNINITIALIZED_VALUE)
  `u` is read without initialization. 
  37.   ma_multi_malloc(0, &v, (size_t) 1, &u, (size_t) 3);
  38.   *v='d';
  39.   *u='o';
         ^
  40.   free(v);  
  41. }


Found 3 issues
[...]

PR #2078, opened 2026-07-13
[pulse] model C variadic out-parameters to fix false MEMORY_LEAK_C

SuryanshSS1011:
Partially addresses #1937 (the `MEMORY_LEAK_C` false positive; the companion `PULSE_UNINITIALIZED_VALUE` false positives are left for follow-up, see Known limitations).

## Problem

Pulse falsely reports `MEMORY_LEAK_C` when memory is allocated and stored through a variadic out-parameter, as in `va_arg(a, char**); *p = malloc(...)`. The `va_arg` result is disconnected from the caller's argument, and `trim_actuals_if_var_arg` drops the variadic actuals at the interprocedural boundary, so there is nothing to reconnect the write to. A plain non-variadic `char**` out-parameter is handled correctly, so this is specific to `va_arg`.

## Fix

This rides the existing specialization mechanism. A new `variadic_actuals` field on `Specialization.Pulse.t` carries the caller's extra actuals, `PulseSpecialization.apply` seeds them into a global bridge array during specialized re-analysis, and the `va_arg` model reads successive elements from it. The read is type-gated to pointer results so value-returning variadics such as a summing `sum(int n, ...)` are unaffected. Regression tests are added and the C pulse suite is otherwise unchanged.

## Known limitations

The companion `PULSE_UNINITIALIZED_VALUE` false positives on the same out-parameters are not addressed here, since those need true caller-actual binding that the specialization's callee-formal-relative heap paths cannot carry, so the affected tests are marked with an `FP_` prefix.

### DIFF
--- a/infer/src/IR/Specialization.ml
+++ b/infer/src/IR/Specialization.ml
@@ -59,13 +59,24 @@ module Pulse = struct
         F.fprintf fmt "}" )
   end
 
-  type t = {aliases: Aliases.t option; dynamic_types: DynamicTypes.t}
+  module VariadicActuals = struct
+    (* the caller's extra (variadic) actual arguments, in order, as heap paths, used to
+       specialize a C variadic callee so that its `va_arg` reads connect to them *)
+    type t = HeapPath.t list [@@deriving equal, compare, hash, sexp, yojson_of]
+
+    let pp fmt actuals =
+      if not (List.is_empty actuals) then
+        F.fprintf fmt "variadic_actuals: [%a] " (Pp.seq ~sep:"; " HeapPath.pp) actuals
+  end
+
+  type t =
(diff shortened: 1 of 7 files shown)


SuryanshSS1011:
@dulmarod @davidpichardie a small nudge on this one. Just rebased onto current main and all checks are green.

The change models C variadic out-parameters so Pulse stops reporting the `MEMORY_LEAK_C` false positive on the `va_arg(a, char**); *p = malloc(...)` idiom, riding the existing specialization mechanism rather than adding a name-based matcher. It is 223 lines across 7 files and scoped to the leak only. The `PULSE_UNINITIALIZED_VALUE` half of #1937 needs caller-actual binding that specialization cannot carry, so those tests stay marked `FP_` and I described what a real fix would take on the issue.

Would love your read on this!

PR #2080, opened 2026-07-22
[merge] avoid unsafe fork(2) on macOS when merging tenvs

jeremydubreil:
The tenv merge worker in `TenvMerger.start` is useing `Unix.fork ()` without a
following `exec`. This is unsafe on macOS and is already checked for the same reason in `Config.unix_fork` to prevent crashing the child with SIGBUS. Adding these changes has been fixing the CI tests on macOS on my local for of Infer.


davidpichardie:
Great! 🥰

davidpichardie:
Great! 🥰

PR #2088, opened 2026-08-04
[pulse] add --pulse-taint-policy-as-issue-type option

jeremydubreil:
Add a `--pulse-taint-policy-as-issue-type` option that, for Pulse taint errors, derives the reported issue type and human-readable bug type from the matching taint policy's name (extracted from the qualifier message) instead of the generic `TAINT_ERROR`. An issue type explicitly set on the taint policy still takes precedence.

This makes reports distinguish the different taint policies (e.g. using self-documented `SQL_INJECTION` or `UNTRUSTED_FILE_CREATION` issue types) rather than collapsing them all into a single issue type. This is particularly useful when there are many policies generating many different kinds of taint flow findings.


dulmarod:
Nice!

PR #2094, opened 2026-08-22
[java] Fix BufferOverrun FP on String.split with non-positive limit

### LINKED ISSUE #1971: FP in Java's BufferOverrun analysis with String.split(String, int) with non-positive limit
Due to an incorrect modelling of the `String.split(String, int)` method, when the limit is `0` or `-1`, the analyzer assumes the resulting array will have a size [limit, 1]; according to [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#split-java.lang.String-int-), in this case the method will either
- behave like the `String.split(String)` overload, when the limit is `0`
- in some cases return one more element than the overload, when the the string ends with a match, leading to an additional empty string not being split from the result; _although I think treating `0` and `-1` the same way, as if the 1 argument overload had been called, would be reasonable_

For example
```java

class Test {
  void method() {
    String s = "a,b";
    String[] split = s.split(",", -1);
    int[] _unused = new int[split.length];
  }
}
```
`split.java:5: error: Alloc May Be Negative,
  Length: [-1, 1].`
Reports `INFERBO_ALLOC_MAY_BE_NEGATIVE` on the integer array allocation.

_I have a possible fix [here](https://github.com/friedbyalice/infer/commit/b4edf14fade78a634ef85eaa228d8578915c5d55), will open a PR after adding tests and some improvement (not sure why it would ever return [-1, 1], covering literals is not enough)_

tanvir-ux:
`String.split(regex, limit)` was modeled by treating `limit` as the malloc size of the result array. For `s.split(",", -1)` that produces `INFERBO_ALLOC_MAY_BE_NEGATIVE` with length `[-1, 1]`, even though the result length is never negative.

Size the result from the receiver (same idea as the one-arg overload). A positive `limit` is only used as an upper bound.

Fixes #1971

### DIFF
--- a/infer/src/bufferoverrun/bufferOverrunModels.ml
+++ b/infer/src/bufferoverrun/bufferOverrunModels.ml
@@ -1638,13 +1638,20 @@ module JavaString = struct
     {exec; check= no_check}
 
 
-  let split_with_limit limit_exp =
+  (** [String.split(regex, limit)]. Non-positive [limit] behaves like [split]. Positive [limit] caps
+      the result length. Allocate from the receiver string; [limit] is not an array size. *)
+  let split_with_limit string_exp limit_exp =
     let exec ({integer_type_widths} as model_env) ~ret mem =
-      let dummy_exp = Exp.zero in
-      let length =
-        Sem.eval integer_type_widths dummy_exp mem |> range_itv_one_max_one_mone |> Dom.Val.of_itv
+      let from_string =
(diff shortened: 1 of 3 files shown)


dulmarod:
ERROR 'Running test: direct java performance test'

tanvir-ux:
> ERROR 'Running test: direct java performance test'

Fixed

PR #2122, opened 2026-09-15
[pulse][taint] make the taint matching loop cheaper

jeremydubreil:
## Context

I tried before to enable the taint models by default but this affected the overall analysis time too
much. These two commits remove the two dominant costs in the matching loop.

## What was slow

`PulseTaintItemMatcher.procedure_matches` runs a `List.filter_map` over the whole matcher list, at
every call site, in every disjunct. Two kinds of work were paid once per matcher instead of once per
call:

1. **Loop-invariant values recomputed inside the closure.** The expensive one is `get_proc_name_s`,
   which formats the whole procedure name with `F.asprintf`. Most matcher kinds never read that
   string, and the default Java config never does: of its 107 matchers (38 sources, 67 sinks, 2
   sanitizers), 90 are `ClassAndMethodNames`, 13 `MethodWithAnnotation`, 2 `ClassWithAnnotation` and
   2 `OverridesOfClassWithAnnotation`, with no regex matchers at all. So the name was formatted 107
   times per call site and discarded every time.

2. **One class hierarchy walk per matcher.** `class_names_match` calls
   `PatternMatch.supertype_exists`, so the 90 `ClassAndMethodNames` matchers walked the same
   hierarchy 90 times, allocating a fresh `Typ.Name.name` string at each visited class.

## The changes

**`use lazy evaluation in the taint matching function`** — everything that depends only on
`proc_name`, `proc_attributes` or `block_passed_to` moves out of the closure, and the
string-formatting values become `lazy` so they are computed at most once, and only if a matcher kind
needs them: `proc_name_s`, its objc-regex variant and `block_passed_to`'s name are now `lazy`;
`class_name`, `method_name`, `is_objc_block`, `source_file` and the `procedure_return_type_match`
closure are computed once. A redundant `Procname.get_class_type_name` in the
`OverridesOfClassWithAnnotation` branch is dropped in favour of the hoisted `class_name`.

**`walk the class hierarchy once per call instead of once per matcher`** — `supertype_names` collects
the names of the supertypes into an `IString.Set` with a single `Tenv.fold_supers`, behind a `lazy`
so it is only paid when a matcher actually tests class names. Each matcher then does a set lookup per
name instead of a hierarchy walk. Only classes present in the type environment are added, which is
exactly what `Tenv.mem_supers` applies its predicate to, so matchers keep ignoring classes missing
from the tenv. `ClassAndMethodNames` also tests the method name first now: that is a list membership
on an already-computed string and it rejects almost every matcher before the class names are looked
at.

Both commits keep the semantics: no condition changes meaning, only its cost and, in the last case,
its evaluation order.

## Measurements

Native binaries built from the same tree at `main` and at this branch, both with the expanded
server-side Java taint config installed so the models load by default. Each run analyses a pre-built
capture DB with `infer analyze --pulse-max-heap 100000000` on 16 workers; two runs per configuration,
interleaved. CPU is user+sys, the metric that behaves sanely across a forked worker pool.

| Project | Config | CPU (s) | Wall (s) | taint findings |
| --- | --- | --- | --- | --- |
| Nacos | no Java taint models | 919 | 62.5 | 0 |
| Nacos | `main` | 1454 | 96.8 | 8 |
| Nacos | **this branch** | **958** | **64.8** | 8 |
| BenchmarkJava | no Java taint models | 695 | 45.7 | 0 |
| BenchmarkJava | `main` | 1208 | 80.3 | 2744 |
| BenchmarkJava | **this branch** | **774** | **51.1** | 2744 |
| WebGoat | no Java taint models | 25.0 | 2.0 | 0 |
| WebGoat | `main` | 47.0 | 7.8 | 37 |
| WebGoat | **this branch** | **34.6** | **7.0** | 37 |

That is **−34.1%** analysis CPU on Nacos, **−36.0%** on BenchmarkJava and **−26.4%** on WebGoat.
Put another way, the price of enabling the models drops from +58% to +4.3% on Nacos, from +74% to
+11% on BenchmarkJava and from +88% to +38% on WebGoat. Run-to-run spread was at most 3.5%, well
below every gap reported here.

WebGoat keeps the largest relative overhead but the smallest absolute one (+9.6 s CPU): it is a
16 kLOC project, so the part of the cost that scales with the number of flows that actually fire —
37 findings in very little code — dominates what is left.

Behaviour is unchanged: `report.txt` is byte-identical between `main` and this branch on all three
projects (Nacos 1616 lines, BenchmarkJava 26143, WebGoat 356), and `direct_sil_pulse_test` and
`direct_java_pulse_test` pass with no expected-output changes.

## Caveats

These are `BUILD_MODE=dev` builds, so the absolute CPU figures run high; the ratios are the
meaningful part and every binary was built identically. Peak RSS is unchanged by these commits
(Nacos 1051 → 995 MiB, BenchmarkJava 315 → 316 MiB, WebGoat 857 → 858 MiB), but it does still grow
when the models are enabled (from 464, 289 and 117 MiB respectively). That growth comes from the
matchers that actually fire, so it is a separate issue from this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dulmarod:
That's great!

